PR #106, opened 2023-11-15
Segmentation (no CLI)

edyoshikun:
This PR allows one to use Cellpose segmentation with and without slurmkit. I added a yml config that points to the nucleus and membrane models and a function that performs the segmentation on `czyx_arrays`.

- [ ] There is no current CLI call for this, but we can set it up if needed. 
- [ ]  Tests for the functions


ieivanov:
@edyoshikun will move the segmentation workflow to VisCy after we create ~iphub~

PR #107, opened 2023-11-18
function to merge multiple stores 

edyoshikun:
This PR adds a dataset merging function. This is currently not a CLI, but it can be made if we see fit. This function is currently is used in a slurmkit script that is ran after stabilizing. Additionally this `utils.merge_datasets()` function can be used to copy new zarr stores in parallel, which is certainly faster than running `cp  source/path target/path`.


edyoshikun:
closing this in favor of #145

PR #129, opened 2024-02-01
GPU peak detection with torch

ziw-liu:
Re-implemented an approximate version of `skimage.feature.peak_local_max` with torch to run on GPU. On the mantis workstation (P4000 with 8GB VRAM) it takes ~4s to compute.


ziw-liu:
Most of the runtime is spent on copying the ~3 GB test image through PCIe...

PR #134, opened 2024-02-22
Fix t_idx and additional channels

edyoshikun:
WIP: fixes the t_idx issue from ....

### DIFF
--- a/mantis/analysis/register.py
+++ b/mantis/analysis/register.py
@@ -267,7 +267,7 @@ def find_lir(registered_zyx: np.ndarray, plot: bool = False) -> Tuple:
     # Find the lir Z
     zyx_shape = registered_zyx.shape
     registered_zx_bool = registered_zyx.transpose((2, 0, 1)) > 0
-    registered_zx_bool = registered_zx_bool[zyx_shape[0] // 2].copy()
+    registered_zx_bool = registered_zx_bool[zyx_shape[2] // 2].copy()
     rectangle_coords_zx = lir.lir(registered_zx_bool)
     x = rectangle_coords_zx[0]
     z = rectangle_coords_zx[1]
--- a/mantis/analysis/register.py
+++ b/mantis/analysis/register.py
@@ -264,10 +264,12 @@ def find_lir(registered_zyx: np.ndarray, plot: bool = False) -> Tuple:
     X_slice = slice(rectangle_xy.min(axis=0)[0], rectangle_xy.max(axis=0)[0])
     Y_slice = slice(rectangle_xy.min(axis=0)[1], rectangle_xy.max(axis=0)[1])
 
-    # Find the lir Z
+    # NOTE: this method assumes the center of the image is representative of the center of the object to estimate the LIR in Z
+    # Find the lir Z using ZX
     zyx_shape = registered_zyx.shape
-    registered_zx_bool = registered_zyx.transpose((2, 0, 1)) > 0
-    registered_zx_bool = registered_zx_bool[zyx_shape[2] // 2].copy()
+    registered_zxy_bool = registered_zyx.transpose((2, 0, 1)) > 0
+    # Take middle X-slice to find the LIR for Z.
+    registered_zx_bool = registered_zxy_bool[zyx_shape[-1] // 2].copy()
     rectangle_coords_zx = lir.lir(registered_zx_bool)
     x = rectangle_coords_zx[0]
     z = rectangle_coords_zx[1]
--- a/mantis/cli/apply_affine.py
+++ b/mantis/cli/apply_affine.py
@@ -75,7 +75,7 @@ def apply_affine(
 
     with open_ome_zarr(target_position_dirpaths[0]) as target_dataset:
         target_channel_names = target_dataset.channel_names
-        target_original_shape_zyx = target_dataset.data.shape[-3:]
+        target_shape_zyx = target_dataset.data.shape[-3:]
 
     click.echo('\nREGISTRATION PARAMETERS:')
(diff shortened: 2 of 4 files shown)


ieivanov:
@edyoshikun I am looking at the `utils.py` [diff](https://github.com/czbiohub-sf/shrimPy/pull/134/files#diff-1093b95a67b463e180ba925108ceedc9cb8eb32d137152e85b5468b02c94ed62) and it seems you haven't started from the latest `main` branch? Can you check on this and update the code as needed so it's easier to see the changes you introduced?

edyoshikun:
@ieivanov updated.

mantis/analysis/register.py
```diff
@@ -267,7 +267,7 @@ def find_lir(registered_zyx: np.ndarray, plot: bool = False) -> Tuple:
     # Find the lir Z
     zyx_shape = registered_zyx.shape
     registered_zx_bool = registered_zyx.transpose((2, 0, 1)) > 0
-    registered_zx_bool = registered_zx_bool[zyx_shape[0] // 2].copy()
+    registered_zx_bool = registered_zx_bool[zyx_shape[2] // 2].copy()
```
ieivanov, line 270: Can you explain what is happening here? `registered_zyx.transpose((2, 0, 1))` should result in a 3D array rather than a 2D array as the `zx_bool` label suggests. Could you index `zyx_shape` from the back, e.g. `zyx_shape[-2]` such that it's clear if you're taking the X_size or Z_size?

mantis/cli/apply_affine.py
```diff
@@ -108,7 +108,8 @@ def apply_affine(
         )
         # Overwrite the previous target shape
         Z_target, Y_target, X_target = cropped_target_shape_zyx[-3:]
-        click.echo(f'Shape of cropped output dataset: {target_shape_zyx}\n')
+        cropped_target_shape_zyx = Z_target, Y_target, X_target
```
ieivanov, line 111: This doesn't make sense to me. Is `cropped_target_shape_zyx` more than 3D such that you only need to take the last 3 dims? These two lines look redundant.

mantis/cli/estimate_affine.py
```diff
@@ -33,7 +34,15 @@
 @source_position_dirpaths()
 @target_position_dirpaths()
 @output_filepath()
-def estimate_affine(source_position_dirpaths, target_position_dirpaths, output_filepath):
+@click.option(
+    "--similarity-flag",
```
ieivanov, line 38: This could be simply called `--similarity`. Not a big deal.

mantis/cli/estimate_affine.py
```diff
@@ -72,35 +82,39 @@ def estimate_affine(source_position_dirpaths, target_position_dirpaths, output_f
...
-            - FOCUS_SLICE_ROI_WIDTH : source_channel_X // 2
-            + FOCUS_SLICE_ROI_WIDTH,
-        ],
-        NA_det=NA_DETECTION_SOURCE,
-        lambda_ill=WAVELENGTH_EMISSION_SOURCE_CHANNEL,
-        pixel_size=source_channel_voxel_size[-1],
-    )
+    if source_channel_Z < 2 or target_channel_Z < 2:
+        focus_source_channel_idx = 1
+        focus_target_channel_idx = 1
```
ieivanov, line 87: This will break when registering 2D images with `shape[-3] = 1`. `focus_from_transverse_band` already returns `focus_idx = 0` for 2D images. What was the reason for making this change?

mantis/cli/estimate_affine.py
```diff
@@ -276,24 +290,36 @@ def lambda_callback(layer, event):
...
+            (
+                z_scale_translate_matrix,
+                np.insert(yx_points_transformation_matrix, 0, 0, axis=1),
+            )
+        )  # Insert 0 in the third entry of each row
+
+        scaling_affine = get_3D_rescaling_matrix(
+            (1, target_channel_Y, target_channel_X),
+            (scaling_factor_z, scaling_factor_yx, scaling_factor_yx),
+        )
```
ieivanov, line 320: It looks like this `scaling_affine` is not used?

mantis/cli/utils.py
ieivanov, line 1: This file has a huge number of changes, it looks like functions may have been reordered. @edyoshikun in your mantis dev environment, can you please install the same versions of the formatting libraries (black, flake8, isort, etc.) as in the `pyproject.toml` file? It's also good to install pre-commit as described in https://github.com/czbiohub-sf/shrimPy/blob/main/CONTRIBUTING.md#setting-up-a-development-environment

ieivanov (changes requested):

mantis/cli/estimate_affine.py
ieivanov: It may also be good to rename these variables to `source_channel_focus_idx` and `target_channel_focus_idx`. For a minute I though this is a channel index rather than Z index

mantis/cli/apply_affine.py
edyoshikun: Z_target,Y_target,X_target were just making sure the target variables were replaced. They need to be updated for the `create_empty_hcs_zarr` but these can be changed. 

I'll refactor the code so that we use less variables across the code.

mantis/analysis/register.py
edyoshikun: changed. This is to estimate the Z- LIR since we do rotate the volume in Z through the optimizer. Here, we take the middle of the ZX projection across X as sample to estimate LIR in Z.

ieivanov (approved):

PR #135, opened 2024-03-28
Methods to stitch a position grid array

ieivanov:
TODO:
 - [ ] Convert `compute_image_translation` to CLI and save in config file
 - [ ] Test with non-square image and non-square grids


ieivanov:
Replaced by #137

PR #137, opened 2024-04-29
Methods to stitch a position grid

ieivanov:
She is ready!

This PR adds methods to stitch a grid of positions. Stitch parameters are computed with `>>> mantis estimate-stitch`. Images are then stitched with `>>> mantis stitch`. Image registration parameters are computed on pairs of images using phase cross-correlation; the pipeline can be extended to include more complex transformations. The image mosaic is compiled using "average" blending; other blending modes may be implemented later.

See examples in `/hpc/projects/intracellular_dashboard/ops/2024_03_05_registration_test/live/3-stitch/` and `/hpc/projects/intracellular_dashboard/ops/2024_04_11_Manual_HELA/1-stitch/v3/`

Replaces #135 
See https://github.com/czbiohub-sf/shrimPy/pull/135#issuecomment-2038005161

TODO:
 - [x] Convert `compute_image_translation` to CLI and save in config file
 - [x] Test with non-square image and non-square grids


mantis/cli/stitch.py
```diff
@@ -0,0 +1,197 @@
...
+@input_position_dirpaths()
+@output_dirpath()
+@config_filepath()
+@click.option(
+    "--temp-path",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True),
+    default='/hpc/scratch/group.comp.micro/',
+    help="Path to temporary directory, ideally with fast read/write speeds.",
+)
+@click.option("--slurm", "-s", is_flag=True, help="Run stitching on SLURM")
```
talonchandler, line 36:
```suggestion
```
I'll suggest deleting this and the `if not slurm` block, since this option isn't really optional.

mantis/cli/stitch.py
```diff
@@ -0,0 +1,197 @@
...
+    temp_path: str,
+    slurm: bool,
+) -> None:
+    """
+    Stitch positions in wells of a zarr store using a configuration file generated by estimate-stitch.
+
+    >>> mantis stitch -i ./input.zarr/*/*/* -c ./stitch_params.yml -o ./output.zarr --slurm
+    """
+    if not slurm:
+        raise NotImplementedError("Only SLURM mode is supported.")
```
talonchandler, line 50:
```suggestion
```

mantis/cli/stitch.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+@click.command()
+@input_position_dirpaths()
+@output_dirpath()
+@config_filepath()
+@click.option(
+    "--temp-path",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True),
+    default='/hpc/scratch/group.comp.micro/',
```
talonchandler, line 33: Hmmm I didn't have default write access here. Do you know if there's an everyone-accessible `/tmp/`? Or could the default be `./`?

Doesn't need to be solved now...for now I'll suggest adding `--temp-path` to the docstring.

mantis/cli/stitch.py
```diff
@@ -0,0 +1,197 @@
...
+    input_position_dirpaths: list[Path],
+    output_dirpath: str,
+    config_filepath: str,
+    temp_path: str,
+    slurm: bool,
+) -> None:
+    """
+    Stitch positions in wells of a zarr store using a configuration file generated by estimate-stitch.
+
+    >>> mantis stitch -i ./input.zarr/*/*/* -c ./stitch_params.yml -o ./output.zarr --slurm
```
talonchandler, line 47:
```suggestion
    >>> mantis stitch -i ./input.zarr/*/*/* -c ./stitch_params.yml -o ./output.zarr --temp-dir ./
```

talonchandler (approved):
Big job! Looks great. I like the slurm job dependencies. 

I'm in the middle of recreating your `2024_03_05_registration_test/live/3-stitch/` (~20% through the final step filling the last store) and everything's going smoothly. 

Here are the small snags I've run into:
- `--slurm` parameter tripped me up on the `stitch` call. Seems required, so I suggest dropping it for now?
- the default `--temp-dir` wasn't writeable for me. Default to `./`?
- When I open the stitched zarr in napari I get `GL_MAX_TEXTURE_SIZE 16384 in at least one axis and will be downsampled.`, and the viewable resolution is noticeably worse than the originals. Do we need pyramids? Another trick? I don't think this needs to block this merge, though.

mantis/cli/stitch.py
ieivanov: That's quite strange - I thought all of CompMicro should have access to `/hpc/scratch/group.comp.micro/`. Maybe check with IT? It's possible that you were trying to overwrite one of my temp files?

It's important that all jobs have access to this directory, that's why `/tmp` on our HPC is not a good choice since that space is local and different per node. As you suggest, `./` may be a better more general default, with a suggestion to use `/hpc/scratch/group.comp.micro/` in the docstring

ieivanov:
Thanks for your review @talonchandler! I implemented your suggestions for `--temp-path` and `--slurm` options. Let's merge this and we'll fix other issues through smaller PRs.

>When I open the stitched zarr in napari I get GL_MAX_TEXTURE_SIZE 16384 in at least one axis and will be downsampled., and the viewable resolution is noticeably worse than the originals. Do we need pyramids? Another trick? I don't think this needs to block this merge, though.

That's true - so far this only happens for our largest (OPS) datasets and it hasn't bothered me much. It'll be nice to have a proper fix, which I think is implementing pyramids. This is not a blocker for now - analysis still happens on a FOV basis and detailed inspection can be done on smaller ROIs by loading a selection of the data through python.

mantis/cli/stitch.py
talonchandler: Aha! My mistake. 

I just looked closer and found that you own `/TEMP_A3_600k_38x38_1.zarr`, so when I copied/reran your scripts I got permissions errors from that directory, not `/hpc/scratch/group.com.micro/`.

talonchandler:
LGTM! Thanks @ieivanov.

PR #138, opened 2024-05-15
Acquisition engine fixes

ieivanov:
This PR expand the O3 scan range during autofocus and adds a check that the number of sequenced events is less than the number supported by the TriggerScope for a single sequence.


edyoshikun (approved):
LGTM!

PR #140, opened 2024-06-10
Slurmkit copy n paste

edyoshikun:
Example slurm script that I use for copy pasting and cropping data.


edyoshikun:
Check #145 . This function might be more useful.

PR #145, opened 2024-07-02
CLI for concatenating zarr stores

edyoshikun:
*NOTE: this function might have to be ported to iohub or iphub later*

This PR adds a CLI for concatenating zarr stores. The main purpose is to merge multiples stores similar to append_channels. 

Additionally, since this relies on the `copy_n_paste_czyx()` function to crop the arrays, one can pass cropping parameters for T,Z,Y, and X in the config file.

This helps create toy datasets and merge the datasets at the end of our pipelines and assumes that all the datasets have the same folder structure. 

Edit:
- I made the cropping optional (i.e `all`) by default. The config file by still expose the variables for users who want to do the cropping.


mantis/analysis/AnalysisSettings.py
```diff
@@ -69,6 +69,37 @@ def check_affine_transform(cls, v):
         return v
 
 
+class ConcatenateSettings(MyBaseModel):
+    concat_data_paths: list[str]
+    time_indices: Union[NonNegativeInt, list[NonNegativeInt], Literal["all"]] = "all"
```
ziw-liu, line 74: Negative integers are useful too:

```py
assert [1, 2, 3][slice(-2, -1)] == [2]
```

edyoshikun: Ah yes, for time it's easier. For the slicing... I think it will break my code because of the way we compute the slicing indeces
(i.e Z_slice.stop-Z_slice.start)

ziw-liu: What about:

```py
assert abs((-2) - (-1)) == 1
```

talonchandler (commented):
I will vote for simplifying this PR to concatenation without cropping. 

I understand that there are many shared parts of concatenation and cropping and that you're trying cut down on I/O here, but I value modularity and clearly named CLI functions over I/O concerns, which can be solved later during pipelining. 

I will also suggest `concatenate` over `concatenate_datasets`.

mantis/cli/concat_datasets.py
```diff
@@ -0,0 +1,161 @@
...
+    default=1,
+    help="Number of parallel processes",
+    required=False,
+    type=int,
+)
+def concatenate_datasets(config_filepath: str, output_dirpath: str, num_processes: int):
+    """
+    Concatenate datasets
+
+    >> mantis concat_datasets -c ./concat.yml -o ./output_concat.zarr -j 8
```
talonchandler, line 81:
```suggestion
    >> mantis concatenate-datasets -c ./concat.yml -o ./output_concat.zarr -j 8
```

mantis/analysis/settings/example_concatenate_datasets.yml
```diff
@@ -0,0 +1,6 @@
+concat_data_paths: ["/path/to/data.zarr/*/*/*","/path/to/data2.zarr/*/*/*"]
+time_indices: all
+channel_names: [all,all]
+X_slice: [0, 100]
+Y_slice: [0, 100]
+Z_slice: [0, 100]
```
talonchandler, line 6: If we decide to keep these, an `all` setting would be helpful.

talonchandler (commented):
I needed this utility today and it worked well! Thanks @edyoshikun!

mantis/analysis/settings/example_concatenate_datasets.yml
edyoshikun: I made them optional by making them 'all'. One can remove them from the example yml file for now if needed.

mantis/analysis/AnalysisSettings.py
```diff
@@ -69,6 +69,37 @@ def check_affine_transform(cls, v):
         return v
 
 
+class ConcatenateSettings(MyBaseModel):
+    concat_data_paths: list[str]
+    time_indices: Union[int, list[int], Literal["all"]] = "all"
+    channel_names: list[Union[str, list[str]]]
+    X_slice: Union[list[int], Literal["all"]] = 'all'
+    Y_slice: Union[list[int], Literal["all"]] = 'all'
+    Z_slice: Union[list[int], Literal["all"]] = 'all'
```
ziw-liu, line 78:
```suggestion
    time_indices: Union[int, list[int], Literal["all"]] = "all"
    channel_names: list[Union[str, list[str]]]
    X_slice: Union[list[int], Literal["all"]] = "all"
    Y_slice: Union[list[int], Literal["all"]] = "all"
    Z_slice: Union[list[int], Literal["all"]] = "all"
```

mantis/analysis/settings/example_concatenate_datasets.yml
```diff
@@ -0,0 +1,6 @@
+concat_data_paths: ["/path/to/data.zarr/*/*/*","/path/to/data2.zarr/*/*/*"]
+time_indices: all
+channel_names: [all,all]
+X_slice: 'all'
+Y_slice: 'all'
+Z_slice: 'all'
```
ziw-liu, line 6: Inconsistent usage of quotes.

ziw-liu:
What is the testing requirement on shrimPy?

edyoshikun:
I don't think we had a strict requirement, however, I will add a test function that runs the CLI. 
The config is checked with the pydantic models.

Thanks @ziw-liu

talonchandler (approved):
I've used and tested this, and I'm happy with where it's at. I'm fine with merging and moving w/o tests for now (but please go ahead if you're doing tests anyway @edyoshikun). Thanks!

edyoshikun:
I added the test and made some changes to make sure the config and cli files have consistent naming w.r.t. to the other CLIs.

talonchandler:
Thanks @edyoshikun! LGTM

PR #146, opened 2024-07-02
GPU deskew

talonchandler:
This PR changes the existing `scipy.ndimage` deskew to a GPU-accelerated `monai` deskew. 

--
**Benchmarks:**
Deskewing a 280x600x1372 argolight target takes:

(Deskewing function calls, including CPU->GPU->CPU transfers)
(Most relevant to @ieivanov's live applications)
**Before: 61 s**
**After: 5.1 s**

(Deskewing CLI calls, including IO, imports, and other overhead)
**Before: 82 s**
**After: 49 s**

-- 
**The new and old deskews do not match exactly for two reasons:** 

(1) **MONAI's GPU deskew only supports nearest-neighbor and bilinear interpolation modes**, not the linear spline interpolation that we used previously. I chose to use bilinear interpolation, and on close inspection (screenshots don't show any difference) the only differences I observe are:
- a <1 pixel shift 
- <1% interpolation differences
- minor changes in auto-contrast---interpolation differences can change the min and max, so autocontrast can change. Locking the contrast when comparing generates indistinguishable contrast. 

**@ieivanov I would appreciate your help in scrutinizing a few initial deskews as we onboard this change.**

(2) **MONAI's GPU deskew only supports filling empty values with zeros**, not the background-estimated fill that we used previously. This difference is not important because we almost always clip off the overhang, leaving no values to fill. 

You can take a closer look at my argolight tests here:
`/hpc/projects/comp.micro/mantis/2024_04_23_mantis_alignment/2-deskew/test-monai`

--

**Other notes:**
- if a GPU is unavailable, the new deskew will still succeed (slowly)
- I expect that @ieivanov's scatter-gather strategy for splitting a deskew into GPU-memory-sized chunks will still succeed.

### DIFF
--- a/mantis/analysis/deskew.py
+++ b/mantis/analysis/deskew.py
@@ -118,8 +118,7 @@ def deskew_data(
     px_to_scan_ratio: float,
     keep_overhang: bool,
     average_n_slices: int = 1,
-    order: int = 1,
-    cval: float = None,
+    device='cpu'
 ):
     """Deskews fluorescence data from the mantis microscope
     Parameters
@@ -129,18 +128,20 @@ def deskew_data(
         - axis 0 corresponds to the scanning axis
         - axis 1 corresponds to the "tilted" axis
         - axis 2 corresponds to the axis in the plane of the coverslip
+    ls_angle_deg : float
+        angle of light sheet with respect to the optical axis in degrees
     px_to_scan_ratio : float
         (pixel spacing / scan spacing) in object space
         e.g. if camera pixels = 6.5 um and mag = 1.4*40, then the pixel spacing
         is 6.5/(1.4*40) = 0.116 um. If the scan spacing is 0.3 um, then
         px_to_scan_ratio = 0.116 / 0.3 = 0.386
-    ls_angle_deg : float
-        angle of light sheet with respect to the optical axis in degrees
     keep_overhang : bool
         If true, compute the whole volume within the tilted parallelepiped.
         If false, only compute the deskewed volume within a cuboid region.
     average_n_slices : int, optional
         after deskewing, averages every n slices (default = 1 applies no averaging)
+    device : str, optional
+        torch device to use for computation. Default is 'cpu'.
     Returns
     -------
     deskewed_data : NDArray with ndim == 3
@@ -149,10 +150,7 @@ def deskew_data(
         axis 2 is the X axis, the scanning axis
     """
     # Prepare transforms
-    Z, Y, X = raw_data.shape
-
     ct = np.cos(ls_angle_deg * np.pi / 180)
-
     matrix = np.array(
         [
             [
@@ -170,15 +168,17 @@ def deskew_data(
         raw_data.shape, ls_angle_deg, px_to_scan_ratio, keep_overhang
     )
 
-    # to tensor on GPU
-    if torch.cuda.is_available():
-        raw_data = torch.tensor(raw_data).to("cuda")
+    # convert to tensor on GPU
+    # convert raw_data to int32 if it is uint16
+    if raw_data.dtype == np.uint16:
+        raw_data = raw_data.astype(np.int32)
+    raw_data_tensor = torch.as_tensor(raw_data, device=device)
 
     # Returns callable
     affine_func = Affine(affine=matrix, padding_mode="zeros", image_only=True)
 
     # affine_func accepts CZYX array, so for ZYX input we need [None] and for ZYX output we need [0]
-    deskewed_data = affine_func(raw_data[None], mode="bilinear", spatial_size=output_shape)[0]
+    deskewed_data = affine_func(raw_data_tensor[None], mode="bilinear", spatial_size=output_shape)[0]
 
     # to numpy array on CPU
     deskewed_data = deskewed_data.cpu().numpy()
--- a/mantis/analysis/deskew.py
+++ b/mantis/analysis/deskew.py
@@ -1,7 +1,8 @@
 import numpy as np
-from monai.transforms.spatial.array import Affine
-from monai import transforms
 import torch
+
+from monai import transforms
+from monai.transforms.spatial.array import Affine
 
 
 def _average_n_slices(data, average_window_width=1):
--- a/mantis/analysis/deskew.py
+++ b/mantis/analysis/deskew.py
@@ -142,11 +142,6 @@ def deskew_data(
         If false, only compute the deskewed volume within a cuboid region.
     average_n_slices : int, optional
         after deskewing, averages every n slices (default = 1 applies no averaging)
-    order : int, optional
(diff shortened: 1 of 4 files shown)


mantis/analysis/deskew.py
```diff
@@ -158,34 +161,36 @@ def deskew_data(
...
                 0,
                 px_to_scan_ratio,
-                Z_shift,
+                0,
             ],
-            [-1, 0, 0, Y - 1],
-            [0, -1, 0, X - 1],
+            [-1, 0, 0, 0],
+            [0, -1, 0, 0],
+            [0, 0, 0, 1],
```
talonchandler, line 175: MONAI needs this additional row.

mantis/analysis/deskew.py
```diff
@@ -139,53 +142,47 @@ def deskew_data(
...
-    deskewed_data = scipy.ndimage.affine_transform(
-        raw_data,
-        matrix,
-        output_shape=output_shape,
-        order=order,
-        cval=cval,
-    )
+    # to tensor on GPU
+    if torch.cuda.is_available():
+        raw_data = transforms.ToDevice("cuda")(torch.tensor(raw_data))
```
ziw-liu, line 176: This can be replaced by:

```suggestion
        raw_data = torch.tensor(raw_data).to("cuda")
```

However this only works for one GPU. I think it's better to explicitly pass in the GPU ID, in which case it might make sense to make this a Tensor->Tensor function and handle the device one level up.

talonchandler: Thanks @ziw-liu, that's much better. 

I'm less certain about making this a Tensor->Tensor function, since this would be the first function in shrimpy that uses that pattern. We currently use `process_single_position` to call this (and other) processing functions, so we would need to make `process_single_position` capable of converting to Tensors for some specific functions.    

I think we should handle this question during our next pipelining hackathon. 

Any thoughts @ieivanov?

ieivanov: Fantastic, thank you! I'll review this in detail. On the mantis computer, we do now have two GPUs, but I've set one up for display and the other for compute - the nvidia driver allows for that functionality. I have used `tensor.to('cuda')` calls and they get sent to the right GPU. I'm not so concerned for processing on the mantis computer, but we should check if anything bad happens on bruno. I do think that this function should take and return ndarrays.

edyoshikun: the above should work if you always have a GPU, which seems to be our case. 
If you want to make it agnostic between them:
```
device = torch.device("cuda" if torch.cuda.is_available() else "CPU")
raw_data = torch.tensor(raw_data).to(device)
```

mantis/analysis/deskew.py
```diff
@@ -139,53 +141,47 @@ def deskew_data(
...
+        raw_data = torch.tensor(raw_data).to("cuda")
+
+    # Returns callable
+    affine_func = Affine(affine=matrix, padding_mode="zeros", image_only=True)
+
+    # affine_func accepts CZYX array, so for ZYX input we need [None] and for ZYX output we need [0]
+    deskewed_data = affine_func(raw_data[None], mode="bilinear", spatial_size=output_shape)[0]
+
+    # to numpy array on CPU
+    deskewed_data = deskewed_data.cpu().numpy()
```
edyoshikun, line 184: this looks good to me. you convert to tensor and then spit out a numpy array. This would work fine with the `process_single_position_v2` and the `apply_function_czyx_v2`

mantis/analysis/deskew.py
ieivanov: I think for now I'll add a `device='cpu'` argument to the deskew function such that it executed on CPU by default. This will not break our processing pipelines and as Talon pointed out, the end-to-end gain on the CLI call is smaller. The big win I see is for processing of freshly acquired data still in memory which will happen as a directly python call, possibly with `device='cuda'` argument.

talonchandler: I don't think any additional change is needed for the behavior you're describing @edyoshikun @ieivanov. 

I tested the implementation in this branch on a machine without a GPU and it still runs. My impression is that if you make a `tensor` it will be on the `cpu` by default, and here I'm only transferring to the gpu if `torch.cuda.is_available()`.

ieivanov: Right, but you'll always transfer it to GPU if one if available and that sometimes may not be desirable - e.g. if only a small gpu is available. I'm working on this now

talonchandler: Gotcha, makes sense. Thanks.

ieivanov (approved):
This look good to me, thanks @talonchandler. Let's see if we can get the tests to pass now

ieivanov:
I think this is ready, @talonchandler @edyoshikun could you please take another look?

talonchandler:
LGTM! Thanks for wrangling tests, @ieivanov.

mantis/analysis/deskew.py
```diff
@@ -127,65 +128,62 @@ def deskew_data(
...
-        matrix,
-        output_shape=output_shape,
-        order=order,
-        cval=cval,
-    )
+    # convert to tensor on GPU
+    # convert raw_data to int32 if it is uint16
+    if raw_data.dtype == np.uint16:
+        raw_data = raw_data.astype(np.int32)
+    raw_data_tensor = torch.as_tensor(raw_data, device=device)
```
ziw-liu, line 175: `Affine`'s output is always `float32`, so this extra casting can be saved by always casting to float.
```suggestion
    raw_data_tensor = torch.from_numpy(raw_data.astype(np.float32)).to(device)
```

ieivanov: makes sense, thanks. I double check that this doesn't require any extra GPU memory in practice

edyoshikun (commented):
I tested this with the opencell data and the argolight target. 
The opencell dataset for 1 timepoint 1 channel it took about ~ 3min to apply the deskew. Was that the same experience with you? Somehow I recall this being faster. Was that the same experience you guys had? 

Otherwise this works. The parallelization using the -j flag also worked. I was using the gpu-small-nodes.

talonchandler:
> The opencell dataset for 1 timepoint 1 channel it took about ~ 3min to apply the deskew. Was that the same experience with you? Somehow I recall this being faster. Was that the same experience you guys had?

Thanks for testing @edyoshikun. I only ran before-and-after benchmarks on the Argolight target. 

How large was the opencell volume you tested? Is the ~3 minutes consistent with 49 seconds I clocked for my 280x600x1372 volume.

ieivanov:
We just discussed that the CLI call executed on CPU, as before. `deskew_data` has an option to use GPUs but that's not implemented in the CLI and that's OK because we speed things up with slurm and CPU multiprocessing. The option to deskew data on GPU is most useful for live data reconstruction / visualization, which currently happens through scripts.

PR #147, opened 2024-07-05
Rename `mantis` to `shrimpy`

talonchandler:
This PR is important but not urgent.

For this PR I:
- Did a find and replace of `mantis` with `shrimpy`
- Manually reinstated "mantis microscope" throughout the docs. (mantis is the microscope, shrimpy is the analysis package)
- Manually fixed URLs that used `shrimPy` (and the new `czbiohub-sf`)
- Updated the CLI so that it accepts both `shrimpy` and `shrimPy` (similar to `recOrder`/`recorder`). 
- Updated paths (absolute hpc paths, and relative paths on mantis computer...the acquisition computer will almost certainly need a fresh installation of `shrimPy` to update paths and CLI calls.)


talonchandler:
Closing for now. We'll rename on both sides (`shrimpy` and `biahub`) after the split with this TODO list.

PR #148, opened 2024-07-11
Fix bug in test_estimate_stabilization

ieivanov:
New call to `df.ffill()` outputs floats which need to be converted to ints for indexing


edyoshikun (approved):
thanks for finding this bug. 
These changes make the code more robust by filling in with the mean and ensuring that
`z_idx` is an int. Looks great to me~

edyoshikun:
Let's merge this to fix the remaining CI bugs on other PRs

PR #149, opened 2024-07-11
Restrict to numpy<2

ieivanov:
numpy 2.0.0 breaks tests


talonchandler:
Sounds good to me. 

The issue seems to be through `estimate_stabilization` -> `pystackreg` -> `turboreg`. If the tests pass here, we can open an issue upstream in `pystackreg`.

talonchandler (approved):

PR #156, opened 2024-08-09
Merging autotracker

edyoshikun:
This PR adds autotracker (smart microscopy) tool into shirmPy.

For this PR, I hope to acomplish the following:

- [ ] Checkout the old code and test the different algorithms
- [ ] Write a test to ensure the shifts are going to the expected direction
- [ ] Ensure the algorithms work for our imaging modalities 
- [ ] Integrate it to shrimPy as hook function

### DIFF
--- a/mantis/acquisition/autotracker.py
+++ b/mantis/acquisition/autotracker.py
@@ -1,0 +1,245 @@
+# %%
+
+from typing import Literal
+
+import matplotlib.pyplot as plt
+import numpy as np
+import skimage
+
+from iohub import open_ome_zarr
+from numpy.typing import ArrayLike
+from skimage.exposure import rescale_intensity
+from skimage.feature import match_template
+from skimage.filters import gaussian
+from skimage.measure import label, regionprops
+from skimage.registration import phase_cross_correlation
+
+from mantis import logger
+
+# import cv2
+
+
+# TODO: Make toy datasets for testing
+# TODO: multiotsu should have an offset variable from the center of the image (X,Y)
+# TODO: Check why PCC is not working
+# TODO: write test functions
+# TODO: consider splitting this file into two
+
+
+# %%
+
+
+def calc_weighted_center(labeled_im):
+    """calculates weighted centroid based on the area of the regions
+
+    Parameters
+    ----------
+    labeled_im : ndarray
+        labeled image
+    """
+    regions = sorted(regionprops(labeled_im), key=lambda r: r.area)
+    n_regions = len(regions)
+    centers = []
+    areas = []
+    for i in range(n_regions):
+        centroid = regions[i].centroid
+        centers.append(centroid)
+        area = regions[i].area
+        areas.append(area)
+    areas_norm = np.array(areas) / np.sum(areas)
+    centers = np.array(centers)
+    center_weighted = np.zeros(3)
+    for j in range(3):
+        center_weighted[j] = np.sum(centers[:, j] * areas_norm)
+
+    return center_weighted
+
+
+def multiotsu_centroid(moving: ArrayLike, reference: ArrayLike) -> list:
+    """
+    Multiotsu centroid method for finding the shift between two volumes (ZYX)
+
+    Parameters
+    ----------
+    moving : ndarray
+        moving stack ZYX
+    reference : ndarray
+        reference image stack ZYX
+
+    Returns
+    -------
+    shifts : list
+        list of shifts in z, y, x order
+    """
+    # Process moving image
+    moving = rescale_intensity(moving, in_range='image', out_range=(0, 1.0))
+    stack_blur = gaussian(moving, sigma=5.0)
+    thresh = skimage.filters.threshold_multiotsu(stack_blur)
+    moving = stack_blur > thresh[0]
+    moving = label(moving)
+    # Process reference image
+    reference = rescale_intensity(reference, in_range='image', out_range=(0, 1.0))
+    stack_blur = gaussian(reference, sigma=5.0)
+    thresh = skimage.filters.threshold_multiotsu(stack_blur)
+    reference = stack_blur > thresh[0]
+    reference = label(reference)
+
+    # Get the centroids
+    moving_center = calc_weighted_center(moving)
+    target_center = calc_weighted_center(reference)
+
+    # Find the shifts
+    shifts = moving_center - target_center
+
+    logger.debug(
+        'moving_center (z,y,x): %f,%f,%f', moving_center[0], moving_center[1], moving_center[2]
+    )
+    logger.debug(
+        'target_center (z,y,x): %f,%f,%f', target_center[0], target_center[1], target_center[2]
+    )
+    logger.debug('shifts (z,y,x): %f,%f,%f', shifts[0], shifts[1], shifts[2])
+
+    return shifts
+
+
+_AUTOFOCUS_METHODS = {
+    'pcc': phase_cross_correlation,
+    'tm': match_template,
+    'multiotsu': multiotsu_centroid,
+}
+
+
+# %%
+def main():
+    print('AUTOTRACKER')
+    """
+    Toy dataset
+    translations = [
+        (0, 0, 0),  # Shift for timepoint 0
+        (5, -80, 80),  # Shift for timepoint 1
+        (9, -50, -50),  # Shift for timepoint 2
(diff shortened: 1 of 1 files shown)


mantis/acquisition/autotracker.py
```diff
@@ -330,10 +331,15 @@ def estimate_shift(self, ref_img: ArrayLike, mov_img: ArrayLike, **kwargs) -> np
...
 
-        shifts_zyx = autofocus_method_func(ref_img=ref_img, mov_img=mov_img, **kwargs)
+        shifts_zyx_pix = autofocus_method_func(ref_img=ref_img, mov_img=mov_img, **kwargs)
 
         # shifts_zyx in px to shifts_zyx in um
-        self.shifts_zyx = np.array(shifts_zyx) * self.scale
+        shifts_zyx_um = np.array(shifts_zyx_pix) * self.scale
+
+        # Limit the shifts_zyx, preserving the sign of the shift
+        self.shifts_zyx = np.sign(shifts_zyx_um) * np.minimum(np.abs(shifts_zyx_um), self.shift_limit)
```
edyoshikun, line 340: are we going for all or nothing strategy or if it's off in one dimension it's ok?

mantis/acquisition/autotracker.py
```diff
@@ -330,10 +331,15 @@ def estimate_shift(self, ref_img: ArrayLike, mov_img: ArrayLike, **kwargs) -> np
         if not autofocus_method_func:
             raise ValueError(f'Unknown autofocus method: {self.tracking_method}')
 
-        shifts_zyx = autofocus_method_func(ref_img=ref_img, mov_img=mov_img, **kwargs)
+        shifts_zyx_pix = autofocus_method_func(ref_img=ref_img, mov_img=mov_img, **kwargs)
 
         # shifts_zyx in px to shifts_zyx in um
-        self.shifts_zyx = np.array(shifts_zyx) * self.scale
+        shifts_zyx_um = np.array(shifts_zyx_pix) * self.scale
```
ieivanov, line 337: TODO: make sure `self.scale` is of length 3, and that the Z, Y, X scales are correct. The Z scale changes per experiment and should be fetched or passed from the metadata

mantis/acquisition/autotracker.py
ieivanov: `len(self.shift_limit) = 3`, here we are capping each dimension to its max allowed value

ieivanov:
@tayllatheodoro yesterday we found that the calculated shifts are not being applied in the acquisition. Here are next steps I suggest:
* I'm not 100% sure that the phase_cross_corr function correctly computes the shift. Yesterday we acquired datasets where we manually shifted the image using the XY stage. You can take these data, run them through the phase_cross_corr function and confirm that the computed shift makes sense. Update parameters of the phase_cross_corr function as needed. This can happen offline, i.e. not at the microscope. You can read the acquired data using the iohub `read_images` function: https://github.com/czbiohub-sf/iohub/blob/main/iohub/reader.py#L126
* Refactor `autotracker_hook_fn` such that it computes a shift, calculates the new target position for that field of view, and saves that target position in a file. It should not be updating the position list.
* Refactor `update_position_autotracker` such that it reads the file saved by `autotracker_hook_fn` and updates the position with the latest entry. This can be the same as the current position, or an updated position given the computed shift. Note that `update_position_autotracker` is called for every position and timepoint; `autotracker_hook_fn` runs on a separate thread with a delay because the computation takes time, so a shift may not always be available.
* I don't see a need for having the variable `position_settings.xyz_positions_shift`, I think you can remove it. Here is pseurocode for what update_position_autotracker could look like:
```python
def update_position_autotracker(self):
  for p_idx in len(self.position_settings.xyz_positions):
    # read the new position for the CSV file; there is a separate file for each position
    pos_label = self.position_settings.position_labels[p_idx]
    csv_filepath = output_shift_path / f'{pos_label}.csv'
    updated_position = # read CSV file and grab the XYZ position at the last row
    self.position_settings.xyz_positions[p_idx] = updated_position
```
* The CSV files should be saved under `/log` rather than in the acquisition directory.

I think a lot of this code can be written offline and tested at the end on the microscope. I'll also set up the automaton for you to do testing there.

edyoshikun:
This is the code I used to generate a toy dataset. I took an image, shifted and center cropped to make a `toy_dataset.zarr`

```python
# %%
from iohub import open_ome_zarr
from pathlib import Path
import numpy as np
import napari
import os

os.environ['DISPLAY'] = ":1005"
# %%
viewer = napari.Viewer()
# %%
input_data_path = Path(
    '/hpc/projects/tlg2/virtual_staining/20240801_2dpf_she_h2b_gfp_cldnb_lyn_mscarlet_VS/0-zarr/55hpf_1/55hpf_1.zarr'
)
key = '/0/2/0'
dataset = open_ome_zarr(input_data_path)
T, C, Z, Y, X = dataset[key].data.shape

# %%
crop = 900
t_slice = slice(0, 5)
x_slice = slice(X // 2 - crop // 2, X // 2 + crop // 2)
y_slice = slice(Y // 2 - crop // 2, Y // 2 + crop // 2)
z_slice = slice(35, 66)
channel_idx = [0, 1, 3]
data = dataset[key][0].oindex[t_slice, channel_idx, z_slice, y_slice, x_slice]

# %%
viewer.add_image(data)
# %%
from scipy.ndimage import shift
from tqdm import tqdm

# Example translations for each timepoint (Z, Y, X)
translations = [
    (0, 0, 0),  # Shift for timepoint 0
    (5, -80, 80),  # Shift for timepoint 1
    (9, -50, -50),  # Shift for timepoint 2
    (-5, 30, -60),  # Shift for timepoint 3
    (0, 30, -80),  # Shift for timepoint 4
]

# Calculate crop boundaries
T, C, Z, Y, X = data.shape
crop = 600
z_slice = slice(10, 26)
y_slice = slice(Y // 2 - crop // 2, Y // 2 + crop // 2)
x_slice = slice(X // 2 - crop // 2, X // 2 + crop // 2)

# Applying the translations
shifted_data = np.zeros(
    (
        T,
        C,
        z_slice.stop - z_slice.start,
        y_slice.stop - y_slice.start,
        x_slice.stop - x_slice.start,
    )
)
for t in tqdm(range(T)):
    shifted = np.zeros_like(data[t])
    for c in tqdm(range(len(channel_idx)), leave=False):
        # Apply shifts and crop
        shifted[c] = shift(data[t, c], shift=translations[t])
    # Crop the shifted volume
    shifted_data[t] = shifted[:, z_slice, y_slice, x_slice]
# %%
viewer.add_image(shifted_data)
# %%
output_path = './toy_translate.zarr'
with open_ome_zarr(
    output_path, mode='w', channel_names=['BF', 'nuc', 'mem'], layout='hcs'
) as output_dataset:
    position = output_dataset.create_position('0', '0', '0')
    position.create_image('0', data=shifted_data)

# %%
```

mantis/acquisition/autotracker.py
```diff
@@ -259,7 +260,10 @@ def phase_cross_corr(
     prod = Fimg1 * Fimg2.conj()
     del Fimg1, Fimg2
 
-    norm = np.fmax(np.abs(prod), eps)
+    if normalization:
+        norm = np.fmax(np.abs(prod), eps)
+    else:
+        norm = 1.0
```
talonchandler, line 266: @tayllatheodoro, @ieivanov, and I just chatted about this. I think I'm closer to understanding, but I'm still not 100% sure. 

I made an error in the meeting when I said that `np.fmax` is an element-wise max, so it returns a scalar ([it's docs are a bit confusing](https://numpy.org/doc/2.2/reference/generated/numpy.fmax.html)). `np.fmax` actually compares two arrays element-wise, broadcasting if needed, then returns an array. Pardon me. 

This means that  `prod / norm` is equivalent to [this statement from Wikipedia](https://en.wikipedia.org/wiki/Phase_correlation#Rationale)
Screenshot 2025-04-01 at 2 48 54 PM
and its inverse Fourier transform should contain a shifted delta. So it seems like the implementation matches Wikipedia. 

So I'm still unsure. @tayllatheodoro, can we look at a plot of `corr` together with and without normalization next time you try this?

edyoshikun: @JoOkuma

mattersoflight:
@tayllatheodoro can you please link some key datasets and your last update on confluence on this project?

PR #160, opened 2024-09-18
updating code for pydantic v2

tayllatheodoro:
I only made small changes, where pydantic was called, I updated to pydantic.v1 and I changed the list of dependencies to work with pydantic v2 and current iohub.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 # list package dependencies here
 dependencies = [
   "copylot @ git+https://github.com/czbiohub-sf/coPylot",
-  "iohub==0.1.0",
+  "iohub @ git+https://github.com/czbiohub-sf/iohub.git",
   "matplotlib",
   "napari; 'arm64' in platform_machine",                 # without Qt5 and skimage
   "napari[all]; 'arm64' not in platform_machine",        # with Qt5 and skimage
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ classifiers = [
 # list package dependencies here
 dependencies = [
   "copylot @ git+https://github.com/czbiohub-sf/coPylot",
-  "iohub==0.1.0",
+  "iohub @ git+https://github.com/czbiohub-sf/iohub.git",
```
edyoshikun, line 24: I think Talon is writing something but pin this to `0.2.0a1`

tayllatheodoro: Done

edyoshikun (approved):
LGTM! I was able to call the `mantis -h` and the `iohub -h` without problems. Thanks @tayllatheodoro

talonchandler:
Looks good, thanks @tayllatheodoro! Merging. 

FYI @ieivanov @edyoshikun I made a very short attempt to fully migrate to pydantic `v2` like we did for `biahub`, and I found that  most of the acquisition settings tests failed. To err on the conservative side and avoid breaking mantis acquisitions, let's merge this and I'll let you or Ed upgrade and test the acquisitions settings on the scope as needed.

PR #161, opened 2024-09-20
Depreciation warnining

tayllatheodoro:
The depreciation warning was added as a decorator before each function and class that are migrating to biahub.
To do:

- [x] Tests


ieivanov (commented):
I find that the `warnings.warn()` message is not issued on the terminal as `click` filters our these warnings. Instead of using `warnings.warn()` you should modify the click command decorator to `@click.command(deprecated=True)`. I've tested that this does issue a warning. With this syntax, you don't have an option to add a custom message (e.g. Use the function in `biahub`), but I think that's OK.

mantis/cli/run_acquisition.py
ieivanov, line 1: `run_acquisition` does not need to be deprecated either. This function will stay in `shrimpy`

mantis/tests/test_analysis/test_analysis_settings.py
ieivanov, line 1: As we discussed, we don't need to test the base pydantic models - these are tested much more comprehensively by the pydantic library. What would be good to test (likely in a future PR) is all the validators that we have written, e.g. test that `@validator("chunks_czyx")` does throw a ValueError if you don't provide a list of four integers.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
 # list package dependencies here
 dependencies = [
   "copylot @ git+https://github.com/czbiohub-sf/coPylot",
+  "deprecated",
```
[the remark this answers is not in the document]
tayllatheodoro: ook

mantis/tests/test_analysis/test_analysis_settings.py
tayllatheodoro: should i delete the file?

ieivanov:
> I find that the `warnings.warn()` message is not issued on the terminal as `click` filters our these warnings. Instead of using `warnings.warn()` you should modify the click command decorator to `@click.command(deprecated=True)`. I've tested that this does issue a warning. With this syntax, you don't have an option to add a custom message (e.g. Use the function in `biahub`), but I think that's OK.

@tayllatheodoro I think this comment is still not resolved. Once you've implemented the `@click.command(deprecated=True)` syntax I think we can merge this branch.

tayllatheodoro:
> deprecated=True

Hii, I had understood it was a suggestion, but I just corrected it. Thank you

ieivanov (approved):

PR #163, opened 2024-09-27
Bugfix shift distance

ahillsley:
-I merged main into this branch so the majority of the changes are from that 
- The bug fix is 2 lines in analysis.stitch. I adjusted the estimated shift by the ROI used in the phase crosscorrelation before being applied to the raw position of the tile


ieivanov:
Replaced by https://github.com/czbiohub-sf/biahub/pull/16

PR #171, opened 2025-04-10
Migrate to pymmcore plus

jeskesen:
This PR is to address issue #170 .

Issue #170, opened 2025-04-08
Migrate acquisition engine from Pycro-Manager to pymmcore-plus


jeskesen:
pymmcore-plus wraps the MMCoreAndDevices libraries directly in python, eliminating the need for the inter process communication with the Java runtime of MicroManager (which is required by Pycro-Manager). The MDA (multi-dimension acquisition) engine/data structure provided by pymmcore-plus is a good fit for the shrimPy acquisition engine.

ieivanov:
After we've address the two sub-issues above, I think we can start transitioning the shrimpy codebase to pymmcore. The majority of the codebase references the MicroManager core instance which can be accessed with `mmc = CMMCorePlus.instance()` rather than `mmc = Core()` - that change is simple.

The main work will be in transitioning the [MantisAcquisition.acquire](https://github.com/czbiohub-sf/shrimPy/blob/bc12305303645e57ea91a0b3eeffbf31f8c53940/mantis/acquisition/acq_engine.py#L1048) function. pymmcore already provides helpful examples: https://pymmcore-plus.github.io/pymmcore-plus/examples/pycro-manager-api/, though I'd prefer if we do a clean transition and fully implement the new pymmcore API rather than using their `Acquisition` adapter. The trickier bits may be transitioning the hook functions, though I think we have good examples of how to do that.

Lastly, in [MantisAcquisition.update_position_settings](https://github.com/czbiohub-sf/shrimPy/blob/bc12305303645e57ea91a0b3eeffbf31f8c53940/mantis/acquisition/acq_engine.py#L443) we access the MicroManager GUI (aka Studio) to retrieve the list of positions which we have set up. When we use pymmcore we'll no longer have access to the same GUI. In the new workflow, we'll most likely set up a position list in [napari-micromanager](https://github.com/pymmcore-plus/napari-micromanager) and we'll need to transfer these positions to the mantis acquisition engine.

ieivanov:
This migration is exciting as it should allow us to also more easily view the acquired data as in #12

ieivanov:
Note: pymmcore-plus uses a default `PYMM_BUFFER_SIZE_MB` of 250 MB which is quite low. When using MicroManager we set that to ~5 GB, I'll check and we should set that at startup.

ieivanov:
Instead of creating issues, I'll summarize our discussion today here:

Dependency tree:
* Switch shrimpy to pymmcore-plus (#170) - target June 15
[...]

PR #177, opened 2025-05-13
Migrate acquisition engine to pymmcore-plus

jeskesen:
This PR addresses #170 .

This is a copy of #171, but we've moved the development branch into this repository rather than a fork in the acquire-project.

Since migrating the acquisition engine is largely an all-or-nothing endeavor, this PR tracks the entire project's progress: whereas individual features will be merged into this branch with their own PRs.

### DIFF
--- a/mantis/acquisition/acq_engine.py
+++ b/mantis/acquisition/acq_engine.py
@@ -58,8 +58,6 @@ from mantis.acquisition.hook_functions.image_saved_hook_functions import (
 
 
 # Define constants
-LF_ZMQ_PORT = 4827
-LS_ZMQ_PORT = 5827  # we need to space out port numbers a bit
 LS_POST_READOUT_DELAY = 0.05  # delay before acquiring next frame, in ms
 MCL_STEP_TIME = 1.5  # in ms
 LC_CHANGE_TIME = 20  # in ms
@@ -91,8 +89,6 @@ class BaseChannelSliceAcquisition(object):
         Path to Micro-manager which will be launched in headless mode, by default None
     mm_config_file : str, optional
         Path to config file for the headless acquisition, by default None
-    zmq_port : int, optional
-        ZeroMQ port of the acquisition, by default 4827
     core_log_path : str, optional
         Path where the headless acquisition core logs will be saved, by default ''
     """
@@ -102,7 +98,6 @@ class BaseChannelSliceAcquisition(object):
         enabled: bool = True,
         mm_app_path: str = None,
         mm_config_file: str = None,
-        zmq_port: int = 4827,
         core_log_path: str = '',
     ):
         self.enabled = enabled
@@ -134,8 +129,6 @@ class BaseChannelSliceAcquisition(object):
                     core_log_path=core_log_path,
                     buffer_size_mb=2048,
                 )
-
-            logger.debug(f'Connecting to Micro-Manager on port {zmq_port}')
 
             self.mmc = CMMCorePlus()
             self.mmc.loadSystemConfiguration(mm_config_file)
@@ -400,7 +393,6 @@ class MantisAcquisition(object):
         self.lf_acq = BaseChannelSliceAcquisition(
             enabled=enable_lf_acq,
             mm_config_file=mm_config_file,
-            zmq_port=LF_ZMQ_PORT,
         )
 
         # Connect to MM running LS acq
@@ -408,7 +400,6 @@ class MantisAcquisition(object):
             enabled=enable_ls_acq,
             mm_app_path=mm_app_path,
             mm_config_file=mm_config_file,
-            zmq_port=LS_ZMQ_PORT,
             core_log_path=Path(mm_app_path) / 'CoreLogs' / f'CoreLog{timestamp}_headless.txt',
         )
 
--- a/mantis/acquisition/acq_engine.py
+++ b/mantis/acquisition/acq_engine.py
@@ -136,8 +136,9 @@ class BaseChannelSliceAcquisition(object):
 
             logger.debug(f'Connecting to Micro-Manager on port {zmq_port}')
 
-            self.mmc = CMMCorePlus(mm_config_file) #Core(port=zmq_port)
-
+            self.mmc = CMMCorePlus()
+            self.mmc.loadSystemConfiguration(mm_config_file)
+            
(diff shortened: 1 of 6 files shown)


jeskesen:
### Status 05/12/25
**TLDR**

- acquisition engine runs on pymmcore_plus using demo config
- file output is handled using acquire-zarr directly (no position handling yet).

**Detail**

As has been noted before, I'm trying to make the port as minimally invasive as I can.  Thus, where possible, have just tried to use analogous data structures between pycromanager & pymmcore_plus.  The main data structure conversions are:

- [pycromanager.Core](https://pycro-manager.readthedocs.io/en/latest/apis.html#pycromanager.Core) -> [pymmcore-plus.CMMCorePlus](https://pymmcore-plus.github.io/pymmcore-plus/api/cmmcoreplus/#pymmcore_plus.CMMCorePlus)
- [pycromanager.Acquisition](https://pycro-manager.readthedocs.io/en/latest/apis.html#pycromanager.acquisition.acquisition_superclass.Acquisition) -> None
  - File writing functionality handled by acquire_zarr.ZarrStream
  - The Core runs the sequence of events directly
- [pycromanager.multi_d_acquisition_events](https://pycro-manager.readthedocs.io/en/latest/apis.html#multi-d-acquisition-events) -> [useq.MDASequence](https://pymmcore-plus.github.io/pymmcore-plus/guides/mda_engine/#building-sequences-with-mdasequence) or [Iterable[useq.MDAEvent]](https://pymmcore-plus.github.io/pymmcore-plus/guides/mda_engine/#a-multi-event-sequence)
- [pycromanager.Studio](https://pycro-manager.readthedocs.io/en/latest/apis.html#pycromanager.Studio) -> TBD

The above handled most of the engine but there were a couple of modifications that needed to be made: both related to the looping structure, where the outer time & position loops are handled by ShrimPy, while the inner channel & z loops are handled by the engine, which allows autofocus and autoexposure to happen at each new position. 

Previously, a copy of the acquisition_event instance was made and the copy was then modified to have the correct indices.    That couldn't be exactly replicated using pymmcore-plus because MDAEvent inherits from a read-only class.  To get around this, I took advantage of the fact that one can use a list of MDA Events as if it were a MDASequence, so I simply regenerate the sequence by iterating over it, and appending a new object to a list created with modified constructor args.

mantis/acquisition/acq_engine.py
```diff
@@ -774,7 +823,7 @@ def acquire_ls_defocus_stack(
         galvo_range : Iterable
         config_group : str, optional
         config_name : str, optional
-        use_pycromanager : bool, optional
+        use_pymmcore_plus : bool, optional
```
jeskesen, line 826: @ieivanov, do we still need this boolean?

mantis/acquisition/acq_engine.py
```diff
@@ -1213,41 +1239,54 @@ def acquire(self):
...
+                        pos_name=p_label,
+                        slm_image=event.slm_image,
+                        properties=event.properties,
+                        metadata=event.metadata,
+                        action=event.action,
+                        keep_shutter_open=event.keep_shutter_open,
+                        reset_event_timer=event.reset_event_timer
+                    )
+
+                lf_events = [mda_event_from_mda_sequence(event) for event in lf_cz_events]
```
jeskesen, line 1270: Previously, deep copies were made of the events (why?), and then those were modified in place.  Since an MDAEvent is immutable, my workaround was to create a list of new objects with slightly modified constructor args.

A better way of doing this would be to just create the MDASequence here, inside the loop...  but I wanted to touch the program flow as little as possible for now.

mantis/acquisition/acq_engine.py
```diff
@@ -1267,19 +1306,19 @@ def acquire(self):
                 logger.info(f"Waiting {t_wait/60:.2f} minutes until the next time point")
                 time.sleep(t_wait)
 
-        self._ls_acq_obj.mark_finished()
-        self._lf_acq_obj.mark_finished()
-
+        # JGE: HACK to indicate finished. 
+        # self._ls_acq_obj.mark_finished()
```
jeskesen, line 1310: Commented out during development, needs to be put back in.

See [#180](https://github.com/czbiohub-sf/shrimPy/issues/180)

mantis/cli/run_acquisition.py
```diff
@@ -63,8 +63,7 @@ def run_acquisition(
     )
 
     # isort: on
-
-    demo_run = True if 'demo' in mm_config_filepath else False
+    demo_run = 'demo' in mm_config_filepath.lower()
```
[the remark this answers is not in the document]
jeskesen: fixed in #181

mantis/acquisition/acq_engine.py
ieivanov: This function had two ways to acquire data - (1) using direct calls to `mmc.snapImage()` via `microscope_operations.acquire_defocus_stack` and (2) using the `pycromanager` interface. Using pycromanager added the benefit of doing sequences (aka hardware triggered) Z stack acquisitions in which the camera is set to acquire a set of frames at its max framerate and the Z stage is stepped on every exposure pulse. `microscope_operations.acquire_defocus_stack` on the other hand allows data to be displayed in a MicroManager window, but we don't use that functionality within `acquire_ls_defocus_stack`. All in all, I think we can just switch to using pymmcore-plus acquisitions that store the data in the `data` array (note that here we return the data structure as numoy array). The sequences vs not sequenced acquisition is controlled externally by enabling sequencing within the Z stage properties.

mantis/acquisition/acq_engine.py
ieivanov: No need to try to work too hard to maintain the original program flow if it doesn't make sense. I think it's OK to create the correct `MDAEvent` on every iteration of the loop, rather than creating a set of them at the beginning and modifying them as we go.

If it makes the built-in writers work better, we could also create a sequence of events for all timepoint and positions at the start, and then skip some of them if, for example, O1 autofocus fail. You're welcome to get creative where it makes sense rather than trying to stick too close to our old workflows if that's cumbersome.

ieivanov:
Note: pymmcore-plus uses a default PYMM_BUFFER_SIZE_MB of 250 MB which is quite low. When using MicroManager we set that to ~5 GB, I'll check and we should set that at startup.

Issue #184, opened 2025-05-16
Confirm that we can run LF or LS acquisitions only


ieivanov:
shrimPy is built to run two acquisitions simultaneously. To port to other microscopes such as the iSIM it would be useful to run only one acquisition at a time (since that microscopes doesn't have two arms that operate simultaneously). I started writing shrimpy intending for this use case, but I think I gave up halfway through. We should make sure that we can cleanly acquire only one of the two acquisitions. We may need to switch the naming from LF / LS acquisition to Main / Aux acquisition (or something similar). The Main acquisitions will still take hold of shared hardware such as the XY stage.

Issue #179, opened 2025-05-13
Verify pymmcore-plus functionality on Squid microscope.


jeskesen:
This is the first test of the new engine on real hardware.  It has the same 2 cameras as Mantis, but a different stage.  There is also no DAQ to synchronize acquisition.

ieivanov:
I think currently `shrimpy` is built to require a DAC, but that should not be the case. If a DAQ is not present - or if the user does not request hardware triggering, we should run the two acquisitions asynchronously - i.e. call 

```
lf_mmc.run_mda()
ls_mmc.run_mda()
```

Without worrying much if there is a time different on when they start or how long they take. Implementing this may require a change to the config file, for example adding 

```
time_settings:
  num_timepoints: 3
  time_interval_s: 5
  synchronize_acquisitions: True
```

ieivanov:
We discovered that the Prime BSI Express camera that we hooked up on the Squid microscope is broken and needs to go in for repair. We'll prioritize testing on the mantis microscope, with and without synchronization through the NiDAQ.

jeskesen:
While working on this, I ran across an incompatibility bug between pymmcore-plus and the C++ Triggerscope Device adapter.    See pymmcore-plus/pymmcore-plus#478.

PR #192, opened 2025-07-15
LF only acquisition with Mantis hardware.

jeskesen:
Addresses: https://github.com/czbiohub-sf/shrimPy/issues/179 (partial) & https://github.com/czbiohub-sf/shrimPy/issues/184


mantis/cli/run_acquisition.py
```diff
@@ -38,11 +38,20 @@ def load_settings(raw_settings: dict, settings_key: str, settings_class):
     help='''Path to Micro-manager config file
       which will run the light-sheet acquisition''',
 )
+@click.option(
```
ieivanov, line 41: You can fix this now or later. I would suggest renaming the `--mm-config-filepath` argument to `--ls-config-filepath` and removing `--mm-app-path` since it is no longer used.

jeskesen: Yeah, I totally agree with that.  I just didn't know if now was the right time to change command line args.

mantis/acquisition/acq_engine.py
```diff
@@ -411,6 +438,7 @@ def __init__(
         acquisition_name: str,
         mm_app_path: str,
         mm_config_file: str,
```
ieivanov, line 440: Similarly, we can rename `mm_config_file` to `ls_config_file`

mantis/acquisition/acq_engine.py
```diff
@@ -564,8 +595,11 @@ def update_lf_acquisition_rates(self, lf_exposure_times: list):
             return
 
         # Determine label-free acq timing
-        oryx_framerate = float(self.lf_acq.mmc.getProperty('Oryx', 'Frame Rate'))
+        oryx_framerate = float(
```
ieivanov, line 598: Unfortunately, there is no standard way to get the camera framerate in MicroManager - this will work for Spinnaker cameras which have a `Frame Rate` property exposed through the device adapter. Other camera manufacturers may call it something different. Still. I do agree with the change you are making here as `self.lf_acq.mmc.getCameraDevice()` is more general than `Oryx`. As a step further, maybe we can make a `microscope_operations.get_camera_framerate()` function which will standardize retrieving the camera framerate across the few different camera manufacturers we use at the Biohub?

ieivanov: This is a pretty weak suggestion at this point - we have hardcoded device name at many places throughout the code, we can also clean these up in another PR, I know this is not the point of this PR

mantis/acquisition/acq_engine.py
```diff
@@ -1178,11 +1221,12 @@ def acquire(self):
         #     ls_post_camera_hook_fn = partial(start_daq_counters, [self._ls_z_ctr_task])
         # ls_image_saved_fn = check_ls_acq_finished
 
-        # Generate LF acquisition events
+        # Generate LF MDA
```
ieivanov, line 1224: Do we need to check if these are enabled?

ieivanov (approved):
I think this looks good to me. I agree that is resolves #184. Let's aim to get to a point where we can stress test the pymmcore-plus acquisition engine on the mantis microscope with minimal software polishing to confirm that pymmcore-plus can do everything we need it. This will involve running a typical mantis acquisition (two arms, multiple channels, few hours imaging multiple positions) and an acquisition where we only use the LF arm. I think the only step towards that goal is finding a good data writer, correct?

ieivanov:
Let me know if you need to me merge these changes, I can overwrite the failed tests check

mantis/acquisition/acq_engine.py
jeskesen: I think I'd like to move this functionality inside of each acquisition class eventually... so I'll handle it then.

PR #194, opened 2025-08-06
Autofocus using pymmcore-plus

jeskesen:
This PR incorporates fixes to auto-focus resulting from hardware dev/test.

### DIFF
--- a/mantis/acquisition/acq_engine.py
+++ b/mantis/acquisition/acq_engine.py
@@ -1247,9 +1247,17 @@ class MantisAcquisition(object):
 
                 # autofocus
                 if self.lf_acq.enabled and self.lf_acq.microscope_settings.use_autofocus:
+                    # use current position if not set to > 0 value.
+                    starting_pos = self.position_settings.xyz_positions[p_idx][2]
+                    if starting_pos <= 0:
+                        starting_pos = self.lf_acq.mmc.getPosition(
+                            self.lf_acq.microscope_settings.autofocus_stage
+                        )
+
                     autofocus_success = microscope_operations.autofocus(
                         self.lf_acq.mmc,
                         self.lf_acq.microscope_settings.autofocus_stage,
+                        starting_pos,
                     )
                     if not autofocus_success:
                         # abort acquisition at this time/position index
@@ -1257,7 +1265,15 @@ class MantisAcquisition(object):
                             f'Autofocus failed. Aborting acquisition for timepoint {t_idx} at position {p_label}'
                         )
                         continue
-
+                    else:
+                        self.position_settings.xyz_positions[p_idx][
+                            2
+                        ] = self.lf_acq.mmc.getPosition(
+                            self.lf_acq.microscope_settings.autofocus_stage
+                        )
+                        logger.debug(
+                            f'Autofocus successful. Z position updated to {self.position_settings.xyz_positions[p_idx][2]} at position {p_label}'
+                        )
                 # autoexposure
                 if well_id != previous_well_id:
                     globals.new_well = True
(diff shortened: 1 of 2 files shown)


mantis/acquisition/acq_engine.py
```diff
@@ -1249,9 +1249,7 @@ def acquire(self):
                 if self.lf_acq.enabled and self.lf_acq.microscope_settings.use_autofocus:
                     autofocus_success = microscope_operations.autofocus(
                         self.lf_acq.mmc,
-                        self.lf_acq.mmStudio,
                         self.lf_acq.microscope_settings.autofocus_stage,
-                        self.position_settings.xyz_positions[p_idx][2],
```
ieivanov, line 1254: We should continue passing the intended Z position as a starting point of the autofocus algorithm. Imagine Pos0 being far away in Z from Pos100. After we finish imaging Pos100, we'd move to Pos0 and the Z stage will not be in the right place for autofocus to engage - `go_to_position` has logic not to set a new Z position if autofocus is running. What the autofocus methods does is first check if autofocus is still engaged after moving from Pos100 to Pos0 - the microscope is capable of doing that, but it doesn't always happen, especially when moving between wells; if autofocus is still engages once we arrive at Pos0, then great, if not try to re-engage at the intended (starting) Z position for Pos0. You had a good idea last time that the starting Z position for Pos0 may not be ideal after several hours of imaging - in that case we can log the last good Z position for Pos0 and try engaging autofocus there.

mantis/acquisition/acq_engine.py
```diff
@@ -1438,33 +1436,27 @@ def await_cz_acq_completion(self):
...
-        while (
-            not all(
-                (
-                    self.lf_acq.mmc.isSequenceRunning(),
-                    (not self.ls_acq.enabled or self.ls_acq.mmc.isSequenceRunning()),
-                )
-            )
-            and (time.time() - t_start) < buffer_time
+
+        while lf_acq_aborted := self.lf_acq.mmc.isSequenceRunning() or (
```
ieivanov, line 1442: shrimpy allows you to also acquire data without using sequencing. I think in this case this check won't work. Is there a more general `isAcquisitionRunning`, independent of whether it's sequenced for not?

mantis/acquisition/microscope_operations.py
```diff
@@ -182,7 +182,7 @@ def get_total_num_daq_counter_samples(CtrTask: nidaqmx.Task or list):
     return num_counter_samples
 
 
-def autofocus(mmc, mmStudio, z_stage_name: str, z_position):
+def autofocus(mmc: CMMCorePlus, z_stage_name: str):
```
ieivanov, line 185: See comment above, we should continue passing `z_position`

ieivanov (changes requested):

mantis/acquisition/acq_engine.py
```diff
@@ -711,7 +711,7 @@ def setup_daq(self):
             microscope_operations.setup_daq_counter(
                 self._ls_z_ctr_task,
                 co_channel='cDAQ1/_ctr3',
-                freq=self.ls_acq.slice_settings.acquisition_rate,
+                freq=self.ls_acq.slice_settings.acquisition_rate[0],
```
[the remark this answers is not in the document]
jeskesen: Yes, this looked wrong to me too, which is why I had modified it earlier...  but this this was simply reverting it back to its original behavior on main (as it crashed without an index). https://github.com/czbiohub-sf/shrimPy/blob/6e6cb1cb24893e49549102480f96a9ce797c34a3/mantis/acquisition/acq_engine.py#L601

mantis/acquisition/acq_engine.py
jeskesen: I think for this, one needs to monitor if the thread is running... but monitoring the thread wasn't working for the sequenced case.  I will dig into this some more, but can we not block this particular PR for it?

mantis/acquisition/acq_engine.py
jeskesen: I have updated this as we discussed (I think).

PR #207, opened 2025-10-08
O3-autofocus-merge

jeskesen:
Merge Ivan's changes to O3 refocus into pymmcore-plus development branch.  As a side-effect, this gets O3 working in pymmcore-plus.

### DIFF
--- a/examples/acquisition_settings/demo_acquisition_settings.yaml
+++ b/examples/acquisition_settings/demo_acquisition_settings.yaml
@@ -66,3 +66,57 @@ ls_microscope_settings:
     - ['Camera', 'BitDepth', '12']
   z_sequencing_settings:
     - ['Z', 'UseSequences', 'Yes']
+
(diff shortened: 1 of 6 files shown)


ieivanov (approved):
Thanks, this looks good to me. The O3 autofocus currently is a fragile and evolving feature. I just merged #205 which reflects the latest use case. You're welcome to pull these changes in this PR or open a separate one. A bit more context, we've deprecated the `KinesisPiezoMotor` stage, and we are currently setting `use_pycromanager=False` in `acquire_ls_defocus_stack`, i.e. using mmc directly to fetch images rather than running a pycromanager acquisition. We've found that pycromanager acquisitions would stall infrequently, but often enough to break multi-day acquisitions, and getting images through mmc directly is more robust; this will probably be better with pymmcore-plus. All in all, after the mantis rebuild we discussed we won't be running O3 autofocus as it currently stands, so it's OK if you don't worry much about it at this point.

PR #208, opened 2025-10-16
Zarr V3 HCS output

jeskesen:


mantis/acquisition/AcquisitionSettings.py
```diff
@@ -195,3 +195,172 @@ def __post_init__(self):
...
+    shard_sizes: Dict[str, PositiveInt] = field(
+        default_factory=lambda: {'x': 1, 'y': 1, 'z': 1, 'c': 1}
+    )
+
+    # Compression settings
+    compression_codec: Optional[Literal['blosc', 'gzip', 'lz4', 'zstd']] = None
+    compression_level: Optional[int] = None
+
+    # Store settings
+    store_path: Optional[str] = None
```
jeskesen: This probably needs to be tied to the output path more explicitly.

examples/acquisition_settings/demo_acquisition_settings.yaml
```diff
@@ -66,3 +66,57 @@ ls_microscope_settings:
...
+  multiscale: false
+  max_threads: 2  # limit threads for demo
+  chunk_sizes:  # unit: pixels per chunk
+    x: 64
+    y: 64
+    z: 1
+    c: 1
+    p: 1
+    t: 1
+  shard_sizes:  # unit: chunks per shard
```
ieivanov: in iohub we use the term `shards_ratio` which I think is also what you mean here. Could we use the same terminology in shrimpy and iohub, even if that's different from zarr-python or acquire-zarr. I suggest renaming `chunk_sizes` to `chunks` and `shard_sizes` to `shards_ratio`. https://github.com/czbiohub-sf/iohub/blob/56a818632727d6291d061f98562c392b6c8ca841/iohub/ngff/nodes.py#L783

mantis/acquisition/acq_engine.py
```diff
@@ -278,64 +301,109 @@ def setup(self, output_path: Union[str, os.PathLike] = None):
...
+                    # Create HCS layout with plate and wells
+                    plate = Plate(
+                        name=self.zarr_settings.plate_name,
+                        description=self.zarr_settings.plate_description or "",
+                    )
+
+                    # Create wells from position settings
+                    wells = []
+                    for well_id in set(self.position_settings.well_ids):
+                        well = Well(
```
ieivanov: Let's make the default well `0/0` rather than `A/1`. We can work on improving this parsing in the future, seem fragile for now

PR #247, opened 2026-02-24
Add AcquisitionController

ieivanov:
Add AcquisitionController for autofocus-aware acquisition                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                    
  Summary                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                    
  - Introduces AcquisitionController that separates autofocus decision logic from the MDA engine, using a queue-based event flow and stream.skip() from ome-writers to handle autofocus failures                                                                                                    
  - Simplifies MantisEngine by removing setup_event() and exec_event() overrides — the engine now focuses on hardware configuration only                                                                                                                                                            
  - Extracts create_stream() from acquire() so the stream can be shared with the controller                                                                                                                                                                                                         
  - Updates the CLI to use the controller for autofocus-aware acquisitions

  Motivation

  Previously, MantisEngine overrode exec_event() to pad zarr datasets with zeros when autofocus failed. This was fragile (modified superclass behavior) and couldn't support smart microscopy where acquisition events are modified on-the-fly based on acquired data.

  The new AcquisitionController owns the reactive acquisition loop:
  1. Iterates hardware-sequenced events from engine.event_iterator()
  2. Moves the stage and engages autofocus before each position
  3. On success: queues the event for normal execution by the MDA runner
  4. On failure: calls stream.skip() to advance the zarr writer, leaving zeros (fill value)

  This architecture enables smart microscopy by subclassing on_frame_ready() to inspect acquired data and modify upcoming events.

  Key changes

  acquisition_controller.py (new)
  - Queue-based event feeding to mda.run() on a background thread
  - Uses eventFinished signal from pymmcore-plus for synchronization between the controller thread (skip) and the MDA thread (append)
  - Overridable on_frame_ready() hook for smart microscopy

  mantis_engine.py
  - Removed setup_event() override (reverted to superclass)
  - Removed exec_event() zero-padding override
  - Added create_stream(sequence, data_path) — extracted from acquire()
  - acquire() simplified to stream + mda.run() without autofocus concern

  cli/acquire.py
  - shrimpy acquire mantis now uses AcquisitionController

  Test plan

  - test_acquisition_controller.py — 7 new tests covering basic acquisition, autofocus skip behavior, frame counting, on_frame_ready hook, and mixed success/failure across timepoints and positions
  - test_autofocus_failure_pads_with_zeros — end-to-end test with 3 positions, 5 timepoints, deterministic AF failures, zarr verification via iohub
  - All existing engine and integration tests pass (58 total)
  - Depends on eventFinished signal added to pymmcore-plus (add-event-finished-signal branch)

### DIFF
--- a/.github/workflows/pull-request-ci.yml
+++ b/.github/workflows/pull-request-ci.yml
@@ -31,24 +31,6 @@ jobs:
       with:
         fetch-depth: 0  # Fetch all history for hatch-vcs
 
-    - name: Checkout pymmcore-plus
-      uses: actions/checkout@v4
-      with:
-        repository: ieivanov/pymmcore-plus
-        path: pymmcore-plus
-
-    - name: Checkout ome-writers
-      uses: actions/checkout@v4
-      with:
-        repository: ieivanov/ome-writers
-        path: ome-writers
-
-    - name: Move editable sources to expected paths
-      shell: bash
-      run: |
-        mv pymmcore-plus ../pymmcore-plus
-        mv ome-writers ../ome-writers
-
     - name: Install uv
       uses: astral-sh/setup-uv@v6
 
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,11 @@ PACKAGE_NAME := shrimpy
 install:
 	uv sync
 
+.PHONY: install-dev
+install-dev:
+	uv sync
+	uv pip install -e ../pymmcore-plus -e ../ome-writers
+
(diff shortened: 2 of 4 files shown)


ieivanov:
See https://github.com/pymmcore-plus/pymmcore-plus/pull/560#issuecomment-3955749805

PR #257, opened 2026-03-24
DynaTrack v2.0

### LINKED ISSUE #230: DynaTrack: First working implementation on pymmcore-plus acquisition engine

### LINKED ISSUE #256: Position Update Infrastructure
Add infrastructure for dynamically updating stage positions (X, Y, Z) during multi-position         
  acquisitions. After each position's z-stack is fully acquired, a pluggable updater function receives
   the current positions and returns updated coordinates for that position. Updated coordinates are   
  applied to subsequent events via the position store.                                              

  - For multichannel acquisitions, only the first channel's data will be used by the tracking
  algorithm (future work)                                                                             
  - The updater runs asynchronously in a background thread so it doesn't block acquisition
  - A no-op stub updater is provided by default; the actual neuromast tracking algorithm (DynaTrack)  
  will plug in later

jeskesen:
## Summary

Adds position update infrastructure to `MantisEngine` and a full DynaTrack implementation for tracking zebrafish neuromasts across timepoints. Positions (X/Y/Z) can now be updated between visits via a pluggable `PositionUpdater`, with updates running asynchronously so they don't block acquisition.

## Position update infrastructure (`shrimpy/mantis/position_update.py`)

- `PositionStore` — thread-safe storage of current coordinates per position
- `PositionUpdater` — base class; subclasses implement `update(frames, coords) -> coords`
- `PositionUpdateManager` — runs the updater asynchronously in a single-worker thread pool after each z-stack completes; buffers frames via the `frameReady` callback (filtered by `update_channel`)
- `MantisEngine` integration: `setup_event` overrides event X/Y/Z from the store on each position visit; position updates are scheduled from `event_iterator` for accurate logging
- Enabled via `metadata["mantis"]["position_update"]["enabled"]`

## DynaTrack (`shrimpy/mantis/dynatrack.py`, `dynatrack_preprocessing.py`, `dynatrack_worker.py`)

- `DynaTrackUpdater` — FFT phase cross-correlation against a stored reference stack, computed on GPU (torch). Supports per-axis dampening, shift limits, and tracking intervals.
- Preprocessing pipeline (pluggable): `deskew` (biahub) → `phase` (waveorder) → `vs` (VisCy virtual staining). Configurable via `preprocessing: ['deskew', 'phase', 'vs']`.
- `DynaTrackWorker` — runs preprocessing + PCC in a subprocess with its own GIL and GPU context, communicating with the engine via queues. Avoids interference with the acquisition thread.
- Shifts are appended to `<zarr_store>/dynatrack_log.csv` as they're computed.
- Supports `z_device` for routing Z updates to `ObjectiveZ` (vs. `PiezoZ` used during z-stacks).

## ReplayCamera (`shrimpy/mantis/replay_camera.py`)

A `UniMMCore` `SimpleCameraDevice` that serves frames from a pre-acquired OME-Zarr FOV, enabling offline testing of DynaTrack without hardware. Supports channel switching, timepoint auto-increment, Z-stage tracking, and MDA integration. Loads the dataset into memory at init for fast frame access.

## CLI additions

- `shrimpy gui` — launches `pymmcore-gui`, with optional `--unicore` flag for `UniMMCore` (needed by ReplayCamera)

## Demo config

`config/mda/mantis/dynatrack_demo.yaml` — full end-to-end example (deskew + phase + VS + PCC) paired with `MMConfig_demo_ReplayCamera.cfg`.

## Tests

- `test_position_update.py` — store, updater, manager, engine integration, frame buffering
- `test_dynatrack.py` — config, shift computation, dampening, limits, logging
- `test_replay_camera.py` — channel/Z/timepoint handling, MDA integration

## TODOs

- [x] Switch PCC algorithm to GPU
- [x] Speed up deskew
- [ ] Remove excessive debug logging
- [x] Debug tracking - deskew introduces XY rotation, implement image to stage transformation matrix


jeskesen:
> It would also be really helpful if we can run MantisEngine is a "simulation mode" where it reads in pre-acquired data in exec_event and returns it for downstream processing and saving. This will help us troubleshoot the DynaTrack algorithm. I'll point you to a good dataset to use during testing.

@ieivanov, I have a counter proposal.  Rather than baking a file-based simulation into the engine, I'd like to make ReplayCamera: a UniCore-based Camera Adapter.  It seems more re-usable & less invasive. It can be based on iohub.   What do you think?

ieivanov:
I found that the default circular buffer is only 250 MB. Bumping it to 2048 MB seems to solve the issue, and I see pretty much 0% usage of the circular buffer. Dropping it back to 250 MB allowed me to replicate the original issue, I think that's a good sign that the problem is fixed. Will keep a close eye on it still.

ieivanov:
## Algorithm comparison: `256-position-update-infrastructure` (this PR) vs `dynatrack-image-to-stage-transform` (on-scope-validated)

Compared the DynaTrack algorithm in this PR's branch against the monolithic `dynatrack-image-to-stage-transform` branch the #284–#288 stack was split from. **Bottom line: with the default config (`tracking_method: 'pcc'`, acquisition baselines present) the two are algorithmically identical. `256` is a strict superset — same validated PCC core, plus opt-in methods and one edge-case policy change.**

### Core PCC algorithm — identical in both (verified line-by-line)
- **Reference**: stored at first encounter, re-anchored every `reference_update_interval` (re-anchor returns position unchanged, applies no correction).
- **Anchor**: `baseline = position` (the `_acquired_at` coords the stack was *actually acquired at*); correction = `baseline − shift` (negative feedback).
- **Shift pipeline**: phase cross-correlation → ZYX px → ×`scale_z`/`scale_yx` → `shift_limits` → `dampening` → `image_to_stage_matrix_xyz @ shift` → XYZ. Both use the correct (non-walrus) transform form.
- `256` defaults to `tracking_method: 'pcc'`, so out-of-the-box behaviour matches the validated branch.

### ⚠️ One behavioural change inside the shared PCC path
**Missing-baseline handling in `PositionUpdateManager.on_position_complete`:**

| | Validated branch | `256` (this PR) |
|---|---|---|
| No acquisition baseline recorded | falls back to live store snapshot and **proceeds** | **skips** the correction (unless a re-anchor timepoint) |

This only triggers on the pre-fetch race / event-not-applied case. `256` is the more correct choice (a stale-baseline correction would overshoot), but the validated branch was tested *with* the proceed-on-fallback behaviour — so this is the **one item worth re-confirming on the microscope**, since baselines were observed going missing in the demo engine flow.

### `256` adds (absent in the validated branch) — all opt-in / additive
1. Alternative `tracking_method`s: `multiotsu_center_of_mass`, `multiotsu_pcc` (reference-based on Otsu masks), and **referenceless** `intensity_center_of_mass` / `roi_center_pcc`.
2. A second anchor model (referenceless): target = ROI **geometric centre**, no stored reference, corrects from t=0 — conceptually different from PCC's "return the object to its position in the reference image."
3. `wants_reference_refresh` gating + hardening (referenceless exclusion, geometric-centre COM fallback on blank volumes, config warning for `reference_update_interval` + referenceless).

### Net
The validated PCC algorithm is preserved intact behind `tracking_method='pcc'`; new methods are bolted on without touching the default path. The only change to validated behaviour is the **missing-baseline policy (fallback → skip)** — recommend a quick on-scope confirmation before/with merge.

ieivanov:
**Commits used for the comparison above** (branch tips at time of review):

| Branch | Commit | Subject |
|---|---|---|
| `256-position-update-infrastructure` (this PR) | `9074250` | DynaTrack: multi-Otsu center-of-mass & referenceless ROI tracking (#288) |
| `dynatrack-image-to-stage-transform` (validated) | `64e235d` | Merge remote-tracking branch 'origin/main' into dynatrack-image-to-stage-transform |

Common ancestor (merge-base): `a9bfb66` (Avoid cpu gpu transfer (#266)).

ieivanov:
Superseded by #290

PR #273, opened 2026-04-28
Dynatrack image to stage transform

ieivanov:
## Summary

Makes DynaTrack stage-drift correction converge reliably on long timelapses, and adds support for fluorescence (deskew-only) tracking channels.

**Correction math (validated on-microscope, 9 positions × 5 timepoints → deadband, no runaway):**
- Add a configurable `image_to_stage_matrix_xyz` at the dynatrack config level so the image→stage relationship can be calibrated per scope. Fix the demo matrix from a reflection (det -1) to a proper 90° rotation (det +1).
- Apply the shift as **negative feedback** (`updated = baseline - shift`) so the stage moves opposite the measured drift. The previous `+shift` amplified drift on any axis that cleared the deadband.
- Anchor each correction to the stage coords the stack was **actually acquired at** (captured at first-frame arrival, queried from `mmcore.getPosition` for the focus axis) instead of a `PositionStore` snapshot the runner's event pre-fetch may have already advanced. Eliminates the one-cycle oscillation that diverged on late updates.
- Add `reference_update_interval` (N timepoints; 0 = never). Periodically re-anchors the reference stack and skips the correction for that frame, accepting the current stage position as the new baseline. Validated: 15 h run with a fixed t=0 reference saturated (clipped/oscillating PCC shifts); 1 h run with interval=4 stays bounded (max 3.4 µm, 0 timeouts/OOM).

**Backpressure / memory** (developed on the `drain-position-updates` branch and merged into this PR):
- `PositionUpdateManager.drain_pending()` + `event_iterator` drains pending updates at timepoint boundaries, bounding the `ThreadPoolExecutor` queue. Fixes `MemoryError: Buffer overflowed` from unbounded frame-buffer backlog when the updater is slower than the acquisition rate.
- Trim GPU memory: clone the selected channel (drops a ~1 GB view onto the multi-channel preprocessor output), drop the channels dict, `torch.cuda.empty_cache()` per worker job, skip the redundant raw→GPU copy when a preprocessor is active.

**Fluorescence tracking:**
- Allow tracking on a raw fluorescence channel (e.g. GFP) with deskew as the only preprocessing step — no phase reconstruction or virtual staining required.
- Rename the image-volume channel `raw` → `deskewed` (it holds the deskewed input; old name was misleading).
- Sanitize position names to alphanumeric in `_save_debug_channels` so hyphenated names like `1-Pos0000` no longer crash the OME-Zarr debug save.

[...]

Issue #267, opened 2026-04-22
DynaTrack v2


ieivanov:
Main issue that summarizes ongoing and outstanding work on DynaTrack v2

PR #285, opened 2026-06-16
DynaTrack: anchor shifts to acquired stage coords + trim GPU memory

### LINKED ISSUE #281: DynaTrack: Anchor shifts to acquired stage coords and reduce GPU memory
Part of #267.

## Problem
Two related issues observed on long runs:
1. Corrections were anchored to a `PositionStore` snapshot that the runner's event pre-fetch could have already advanced, causing a one-cycle oscillation that diverged on late updates.
2. Per-worker GPU memory grew unbounded — multi-channel preprocessor output held a ~1 GB view, raw→GPU was being copied even when a preprocessor was active, and no `empty_cache()` between jobs.

## Fix
- Capture the stage coords the stack was actually acquired at (first-frame arrival; `mmcore.getPosition` for the focus axis) and use those as the correction baseline.
- Clone the selected channel (drop the multi-channel preprocessor view), drop the channels dict, `torch.cuda.empty_cache()` per worker job, skip the redundant raw→GPU copy when a preprocessor is active.

## Validation
- On-microscope: oscillation eliminated, GPU memory stable across timepoints.

tayllatheodoro:
Closes #281. Sub-PR of #267. Stacks on #284 (PR-2).

Sub-PR 3/6.

## Summary

Two related fixes on long-running acquisitions:

**Acquisition-baseline anchoring.** Corrections were anchored to a `PositionStore` snapshot that the runner's event pre-fetch could have already advanced, causing a one-cycle oscillation that diverged on late updates. Capture the stage coords the stack was actually acquired at (first-frame arrival; `mmcore.getPosition` for the focus axis) inside `PositionUpdateManager.apply_position_update` and pass them as `position` to the updater. The shift now compensates the drift between the reference and where the stack actually was.

**GPU memory.** Per-worker GPU memory was growing unbounded:
- Store the selected channel as `.clone()` so the stored reference no longer pins the entire multi-channel preprocessor output via a view (~1 GB → ~0.5 GB per position).
- Drop the channels dict + `torch.cuda.empty_cache()` after channel select and once per worker job.
- Skip the redundant raw→GPU copy when a preprocessor is active.

## Test plan

- [x] Regression tests assert the correction uses the acquisition baseline, not an advanced store value (b8c4229).
- [x] No new test regressions vs PR-2 baseline. +3 tests now pass thanks to b8c4229's stale-assertion fixes.

## Commits

Cherry-picked from #273 (preserving original authorship):
- Anchor DynaTrack shifts to acquired stage coords, not store snapshot (Ivan — first iteration with `acquired_at` parameter)
- DynaTrack: anchor shifts to acquisition baseline; trim GPU memory (Taylla — final form: `acquired_at` removed, baseline captured upstream)

Note: the two commits represent the natural evolution. The second supersedes the `acquired_at` parameter approach with a simpler upstream capture, and adds the GPU memory trim.

### DIFF
--- a/shrimpy/mantis/dynatrack.py
+++ b/shrimpy/mantis/dynatrack.py
@@ -531,7 +531,8 @@ class DynaTrackUpdater(PositionUpdater):
         # Add the shift to get the new position in image space
         # Convert the new position in image space to stage position
         # Update the position in stage space
-        if transform_xyz := self._config.image_to_stage_matrix_xyz is not None:
+        transform_xyz = self._config.image_to_stage_matrix_xyz
+        if transform_xyz is not None:
             shift_stage_xyz = np.asarray(transform_xyz) @ shift_image_xyz
             logger.info(
                 f"DynaTrack: applied image-to-stage matrix transform to shift: "
--- a/shrimpy/mantis/dynatrack.py
+++ b/shrimpy/mantis/dynatrack.py
@@ -532,15 +532,15 @@ class DynaTrackUpdater(PositionUpdater):
         # Add the shift to get the new position in image space
         # Convert the new position in image space to stage position
         # Update the position in stage space
-        image_to_stage_matrix_xyz = self._config.image_to_stage_matrix_xyz
-        if image_to_stage_matrix_xyz is None:
-            image_to_stage_matrix_xyz = np.eye(3)
-        shift_stage_xyz = np.asarray(image_to_stage_matrix_xyz) @ shift_image_xyz
-        logger.info(
-            f"DynaTrack: applied image-to-stage matrix transform to shift: "
-            f"image_xyz=({shift_image_xyz[0]:.2f}, {shift_image_xyz[1]:.2f}, {shift_image_xyz[2]:.2f}) um -> "
-            f"stage_xyz=({shift_stage_xyz[0]:.2f}, {shift_stage_xyz[1]:.2f}, {shift_stage_xyz[2]:.2f}) um"
-        )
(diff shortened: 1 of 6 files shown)


ieivanov:
### Follow-up: skip the correction on a missing baseline (was: race-prone store fallback)

Pushed `c3723bc`. `on_position_complete` previously fell back to the **live `PositionStore` snapshot** when no acquisition baseline was recorded for a `(t, p)` stack. Under the runner's event pre-fetch race that snapshot can already reflect a later correction, so the shift would be applied against a baseline *ahead* of where the stack was actually acquired — re-introducing the very overshoot/oscillation this PR's anchoring was added to prevent.

**Change:** skip the correction instead. Because shifts anchor to the fixed t=0 reference (not the previous frame), a skipped correction is **self-healing** — the next timepoint computes against the same reference and re-centers fully. A missing baseline with no store entry is still a no-op; a missing baseline *with* a store entry now logs at `error` (it signals a wiring issue, not a routine state).

**Test note:** this surfaced that the manager/engine unit tests called `on_position_complete` (directly or via `_on_frame_ready`) *without* first going through `apply_position_update`, so they were silently relying on the store fallback. Added a `_record_baseline()` helper that records the baseline first — matching the real flow, where the event iterator always records it before frames complete — and replaced the falls-back-to-store test with one asserting the correction is skipped.

**Unrelated:** `test_positions_shift_across_acquisitions` fails locally (`demo-PFS call failed` → frames skipped → `KeyError`), but it was already red on this branch before this change — a demo-autofocus environment artifact, not a regression here.

ieivanov (approved):

PR #286, opened 2026-06-16
DynaTrack: periodic reference re-anchor for long timelapses

### LINKED ISSUE #270: DynaTrack: Add option to update reference every ~2h hours

tayllatheodoro:
Closes #270. Sub-PR of #267. Stacks on #285 (PR-3).

Sub-PR 4/6.

## Summary

Add `reference_update_interval` (default 0 = never). On a re-anchor timepoint the current stack becomes the new reference and **no** shift correction is applied — the current stage position is accepted as the new baseline, avoiding a jump.

## Motivation

Validated on-microscope: a fixed t=0 reference saturates over a 15 h run (sample develops, phase-correlation against the stale reference returns large/clipped/oscillating shifts). A 1 h test with `reference_update_interval=4` keeps the reference fresh: max shift 3.4 µm, no clipping, 0 timeouts/OOM.

## Test plan

- [x] Unit tests for re-anchor behavior (in ad37858).
- [x] No new test regressions vs PR-3 baseline; PR-4 has 0 failures (down from 2 on PR-3) — ad37858's test-file updates resolve the residual stale assertions.
- [x] On-microscope: 1 h run with `reference_update_interval=4` validated.

## Commits

Cherry-picked from #273 (preserving original authorship):
- DynaTrack: periodic reference re-anchor for long timelapses

### DIFF
--- a/config/mda/mantis/dynatrack_demo.yaml
+++ b/config/mda/mantis/dynatrack_demo.yaml
@@ -65,6 +65,11 @@ metadata:
         scale_yx: 0.1494
         scale_z: 0.174
         tracking_interval: 1
+        # Re-anchor the reference every N timepoints (0 = never; keep the fixed
+        # t=0 reference). On a re-anchor timepoint the current stack becomes the
+        # new reference and no correction is applied. Use for long timelapses
+        # where the sample changes enough that matching a stale reference fails.
+        reference_update_interval: 0
         save_debug: true
         # shift_limits:
         #   z: [0.5, 2.0]
--- a/shrimpy/mantis/dynatrack.py
+++ b/shrimpy/mantis/dynatrack.py
@@ -62,6 +62,14 @@ class DynaTrackConfig:
         "y", "x". Shifts below min are zeroed; shifts above max are clipped.
     tracking_interval : int
         Track every N timepoints (1 = every timepoint).
+    reference_update_interval : int
+        Re-anchor the per-position reference every N timepoints (0 = never,
+        i.e. keep the fixed t=0 reference). On a re-anchor timepoint the
+        current stack becomes the new reference and NO shift correction is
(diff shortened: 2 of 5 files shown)


ieivanov:
### Follow-up: a missing baseline must skip only a correction, not a due re-anchor

Pushed `9267304`. This builds on the acquisition-baseline skip added on the #285 branch. That skip is correct for a **fixed** reference, but with periodic re-anchoring (this PR) a re-anchor timepoint applies **no** correction — it just adopts the current stack as the new reference and returns the position unchanged. So a missing baseline is harmless there, and an unconditional skip would wrongly defer a scheduled refresh by a **full interval** (`% reference_update_interval`).

**Change:** relocate the decision to the updater, the only component present in *both* thread and subprocess modes that knows the re-anchor schedule:
- `PositionUpdater.wants_reference_refresh(timepoint_index) -> bool` (base returns `False`; the manager stays updater-agnostic and just asks `self._updater`).
- `DynaTrackUpdater` overrides it to mirror `update()`'s interval test.
- `on_position_complete`: on a missing baseline, skip only when the updater does **not** want a refresh at this timepoint; otherwise proceed using the live-store snapshot (a no-op baseline on the refresh path).

Added `test_no_baseline_proceeds_on_reference_refresh`; the existing no-baseline test still asserts plain corrections are skipped.

**⚠️ Rebase note:** this branch is **not yet rebased on the updated #285** (it still carried the old store-fallback). When it's rebased, `on_position_complete` and the `test_position_update.py` baseline-helper changes will overlap with #285 — resolve toward **this** branch's version (the conditional skip supersedes #285's unconditional one). I pulled in #285's test helper here so the post-rebase reconciliation is trivial.

ieivanov (approved):

PR #291, opened 2026-07-08
Smart fov selection

y222gu:
## Summary

This adds online FOV selection to the Mantis engine. Instead of imaging every field of view for the whole timelapse, we do a quick pre-scan first, decide which FOVs are worth keeping, and then run the timelapse only on those.

It runs as two acquisitions back to back. The first images each candidate FOV once and scores it with the trained model as the frames come in; the second images only the FOVs that passed. Both are driven from `acquire()`, so there's no separate controller or special CLI path. It's just a normal `shrimpy acquire mantis` run. I developed and tested it offline against a real A549 dataset using the `ReplayCamera`.

## How a run looks

```
RUN 1  pre-scan     all candidates, one channel, single timepoint, full z
                    output is discarded (unless save_prescan)
                    as each FOV's z-stack lands, a worker subprocess does:
                    deskew -> phase -> virtual stain -> project -> cellpose
                    -> features -> tree -> good/bad
       (drain, collect the good FOV names)
RUN 2  timelapse    the good FOVs only, full channels + z, loops = N
                    written to <name>.ome.zarr
```

`time_plan.loops` is just the number of timelapse timepoints. The pre-scan is its own run, so there's no off-by-one to worry about.

## Why two runs instead of one

`core.mda` won't let you nest runs, but running twice in a row is fine, since each `run()` rebuilds its sink and state. That turned out to be much cleaner than the single-run version I started with.

The pre-scan is decision-making, not science data, so it doesn't belong in the output store. When I tried to keep everything in one run with the pre-scan as timepoint 0, it caused real problems. The pre-scan and timelapse were forced to share one z-range and frame shape, every good FOV carried a throwaway extra timepoint, and the bad FOVs were left as a lonely t=0. Getting rid of it inside a single run is genuinely awkward, because pymmcore-plus ties acquiring a frame to writing it, and the sink's `skip()` doesn't carry an event you can reroute. So you can't easily "acquire but don't save" or send t=0 somewhere else without building a custom split sink. Two runs just sidestep all of it. When you do want the pre-scan on disk, `save_prescan` writes it to its own `<name>_prescan.ome.zarr` instead of discarding it.

## How it's put together

The pre-scan is defined as its own MDASequence under `metadata.mantis.fov_selection.prescan_mda`, with its own positions and z-plan. This was a suggestion from review, and it pays off: the pre-scan can use a different (say, single-slice 2D) z-plan than the timelapse, and it leaves room to later loop the pre-scan and wait for FOVs to become good. The top-level `stage_positions` is left empty and filled in at runtime with the FOVs that passed.

- `acquire()` stays thin: build the pre-scan sequence, run it, read back the good FOV names (captured in `teardown_sequence`), build the timelapse on those, run it. If nothing passed, it logs a warning and skips the second run.
- The sequence-building helpers live in `shrimpy/fov_selection/sequences.py`, not the engine, so the engine isn't cluttered with experiment-specific logic.
- `shrimpy/fov_selection/` holds the rest: `manager.py` (the streaming coordinator, which keeps only one decision in flight so a whole plate never sits in memory), `worker.py` (a subprocess, to keep torch/GPU work away from the camera), and `pipeline.py` (the actual per-FOV decision, reusing the offline feature code so the online numbers match training).
- Small optimization: the decision only computes the features the model actually needs, and anything derivable from the mask alone skips the per-object `regionprops` pass. Same numbers, a lot faster.
- Preprocessing (deskew, phase, virtual staining) is shared with DynaTrack in `shrimpy/preprocessing.py` rather than copied.

## Configuration (`config/mda/mantis/fov_online_demo.yaml`)

Candidate FOVs are a useq `WellPlatePlan` (pick wells plus a per-well grid), which gives a proper HCS OME-Zarr on the way out. The `deskew`/`phase`/`virtual_staining` blocks sit under `fov_selection`; pixel size and z step aren't repeated in the config, since they're read from the Core and the z-plan at runtime. Two channel keys are easy to mix up: `fov_selection_channel` is the one channel we acquire and reconstruct, while `fov_selection_channels` (plural) is what the decision is computed on. Flags: `save_prescan` keeps the pre-scan store, and `save_decision` writes per-FOV debug PNGs plus one combined `fov_summary.csv` into a `<name>_fov_debug/` folder.

## HCS field naming

Worth flagging because it's easy to trip over, and it is not replay-specific: it applies equally on the real scope. When the good FOVs are filtered out of the `WellPlatePlan` and rebuilt as explicit positions, useq names them like `B4_0000`. iohub rejects that as a well field path because it isn't alphanumeric. The fix reduces the field name to just the FOV index (`0000`), which matches what ome-writers produces, so the resulting store opens cleanly. This lives in `_filter_good_positions` in the shared acquisition path, independent of the camera.

## Replay camera

`ReplayCamera` reads OME-Zarr (v2/v3) through dask, handles HCS plates, and maps each run position back to the right source FOV by plate row/column and per-well order. That's what lets the exact same engine code run offline. On the real scope the `WellPlatePlan` drives the stage directly and no mapping is needed.

## Validation

Full two-run acquisition on the replay dataset (wells B/3 to B/5). The scores came out cleanly split, with B/3 bad (0.028) and B/4 and B/5 good (1.000), so the timelapse ran only on B/4 and B/5. The resulting store is a proper HCS OME-Zarr of just those FOVs and opens fine in napari/iohub.

## Tests

`test_preprocessing.py`, `test_fov_selection_manager.py`, `test_fov_selection_acquire.py` (the sequence builders and the HCS-preserving position filter), and `test_fov_selection_pipeline.py` (the feature fast path, pinned against the full path). Full suite is at 285 passing.


ieivanov (commented):
Great first draft! A couple of high-level comments:
* It's OK to modify `mantis_engine.py` to integrate FOV selection. Currently, we do one pre-scan followed by timelapse imaging of the selected FOV. In the future, we may continuously evaluate whether a given FOV is still "good" over time. A good design would be to do the the pre-scan in the `setup_sequence` method and check if the FOV is "good" in the `event_iterator`.
* It's best to unify the preprocessing between DynaTrack and FOV selection instead of create a separate copy. In that case, `preprocessing.py` should be refactored under `shrimpy/preprocessing.py`. You should add tests for the features of the preprocessing pipeline that you rely on for fov selection
* It's OK if ReplayCamera only works with uint8/uint16 data. I shared with you the correct starting dataset for this test.
* If data is not saved in the same format as expected, you are likely not including the needed fields under `stage_positions` in the yaml config file. Positions should have `plate_row` and `plate_col` indices to be saved in HCS zarr format.

shrimpy/cli/acquire.py
```diff
@@ -111,7 +111,29 @@ def mantis(
...
+    from useq import MDASequence
+
+    sequence = MDASequence.from_file(mda_config)
+    fov_cfg = (
+        (sequence.metadata.get("mantis", {}) or {}).get("fov_selection")
+        if sequence.metadata
+        else None
+    )
+
+    if fov_cfg and fov_cfg.get("enabled"):
```
ieivanov: It's best to fold the FOV selection algorithm directly in the mantis acquisition engine using the `setup_sequence`, `setup_event`, and `event_iterator` functions

config/mda/mantis/fov_online_demo.yaml
```diff
@@ -0,0 +1,96 @@
...
+      enabled: true
+      input_channel: "BF - Oblique"   # channel fed to reconstruction
+      projection: sum                 # model uses *_vs_sum__* features
+      threshold: 0.5                  # P(good) cutoff
+      pixel_size_um: 0.1133           # true XY pixel size (replay core reports a preset)
+      model_path: /hpc/projects/comp.micro/microscope_dev/smart_fov_selection/fov_selection_output/fov_features/split_analysis/max_accuracy/model_on_three_dataset/final_model_nuclei-vs_membrane-vs.joblib
+
+      # Explicit pre-scan block (single timepoint forced; z_plan omitted -> timelapse's).
+      prescan:
+        channels:
```
ieivanov, line 101: It's best if `prescan` is a valid `MDASequence` rather than inventing a new format. When you parse it, you may chose to throw an error if, for example, the number of timepoints is greater than one. The prescan should include definition of `stage_position` and `z_plan`. `stage_position` should define which FOVs will be search. The main `stage_position` should be empty. `z_plan` may define a single z slice, which will be useful when doing FOV based on fluorescence.

config/mda/mantis/fov_online_demo.yaml
```diff
@@ -0,0 +1,96 @@
...
+    group: Channel
+    exposure: 1.0
+
+# ---- FOV-SELECTION parameters (pre-scan + reconstruction + decision) ----
+metadata:
+  mantis:
+    fov_selection:
+      enabled: true
+      input_channel: "BF - Oblique"   # channel fed to reconstruction
+      projection: sum                 # model uses *_vs_sum__* features
```
ieivanov: projection should be part of the preprocessing, for example `preprocessing: ['deskew', 'phase', 'vs', 'sum_projection']`

config/mda/mantis/fov_online_demo.yaml
```diff
@@ -0,0 +1,96 @@
...
+      prescan:
+        channels:
+          - config: BF - Oblique
+            group: Channel
+            exposure: 1.0
+
+      # Reconstruction: BF -> nuclei/membrane. Mirrors the DynaTrack config; the
+      # VS model is the A549-finetuned VSCyto3D. pixel size / z step are injected
+      # from pixel_size_um above and z_plan.step (single source of truth).
+      reconstruction:
```
ieivanov: When defining params for the reconstruction, match the style in `config/mda/mantis/dynatrack_demo.yaml`. deskew, phase, and virtual_staining params should come directly under `fov_selection`

config/mda/mantis/fov_online_demo.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - config: GFP EX488 EM525-45
+    group: Channel
+    exposure: 1.0
+
+# ---- FOV-SELECTION parameters (pre-scan + reconstruction + decision) ----
+metadata:
+  mantis:
+    fov_selection:
+      enabled: true
+      input_channel: "BF - Oblique"   # channel fed to reconstruction
```
ieivanov: Like in DynaTrack, you should define `input_channel` and `fov_selection_channel`. `fov_selection_channel` may be one of the input channels - in cases where we do FOV selection directly on acquired data or one of the preprocessed channels.

config/mda/mantis/fov_online_demo.yaml
ieivanov: `segmentation` should also be explicitly defined as preprocessing steps. Some FOV selection algorithms (for example based on DynaCLR) may not go through segmentation. You should also include a section on segmentation parameters. Ideally, we can switch between segmentation models (e.g. watershed by cellpose) and tune the segmentation at least to some extend.

config/mda/mantis/fov_online_demo.yaml
```diff
@@ -0,0 +1,96 @@
...
+      projection: sum                 # model uses *_vs_sum__* features
+      threshold: 0.5                  # P(good) cutoff
+      pixel_size_um: 0.1133           # true XY pixel size (replay core reports a preset)
```
ieivanov: You should fetch the pixel size from the Core, like we do in DynaTrack

config/mda/mantis/fov_selection_demo.yaml
```diff
@@ -0,0 +1,93 @@
...
+      enabled: false
+      method: demo-PFS
+      stage: Z
+    # Pseudocode
+    fov_selection:
+      enabled: true
+      model: tree_1
+      target_fov_count: 9
+      # Pre-scan for FOV selection
+      search_mda:
```
ieivanov: This is a good example of a prescan MDA definition

docs/fov_selection_integration_plan.md
```diff
@@ -0,0 +1,94 @@
...
+  disk (~one FOV, a few GB, in RAM). The plate never sits in RAM.
+
+## Reconstruction = COPY DynaTrack's preprocessing (decoupled), not import it
+
+The deskew -> phase -> virtual-staining chain already exists in
+`shrimpy/dynatrack/preprocessing.py` (`build_preprocessor` /
+`_LabelfreePreprocessor`, using biahub `DeskewSettings`, waveorder phase, and
+cytoland VS).
+
+**Decision:** copy it into `shrimpy/fov_selection/` (e.g.
```
ieivanov: We should reuse the DynaTrack processing chain

docs/fov_selection_integration_plan.md
```diff
@@ -0,0 +1,94 @@
...
+`fov_selection/preprocessing.py`) rather than importing from `dynatrack/`, so
+FOV-selection reconstruction can be tuned independently without affecting
+DynaTrack's tracking behavior. The two are separate.
+
+This is not a pure file copy — `preprocessing.py` is coupled to
```
ieivanov: It's OK to modify to some extent how DynaTrack is build for the purpose of easier reuse of its modules

shrimpy/fov_selection/feature_viewer/__main__.py
```diff
@@ -0,0 +1,4 @@
+from .app import main
```
ieivanov, line 1: This feature viewer can be a utility outside of shrimpy; for now it's OK as it is

shrimpy/scripts/predict_fov_goodness.py
```diff
@@ -0,0 +1,354 @@
+"""
```
ieivanov, line 1: It looks like some of these are independent scripts? If so, it's best that they live under `shrimpy/scripts/`

ieivanov (commented):
Thanks for the updates! Two sequential acquisition runs make a lot of sense. Feel free to resolve comments above you've addressed already.

shrimpy/mantis/mantis_engine.py
```diff
@@ -182,31 +207,46 @@ def setup_sequence(self, sequence: MDASequence) -> SummaryMetaV1 | None:
             )
             self._dynatrack.start(zyx_shape=zyx_shape, log_file_path=_find_shrimpy_log_file())
 
+        # FOV selection runs its reconstruction in a worker subprocess for the
+        # same torch/GPU isolation reason; start it after the ROI is applied so
+        # the acquired frame shape (used to build the transfer function) is known.
+        if self._fov is not None:
```
ieivanov, line 237: You could pull out the zyx_shape = (...) logic such that it can be reused by both dynatrack and fov selection

shrimpy/mantis/mantis_engine.py
```diff
@@ -472,26 +527,191 @@ def acquire(
...
-                root_path=data_path, compression="blosc-zstd", format="acquire-zarr"
-            ),
+            output=out_settings,
             dimension_overrides={"z": {"chunk_size": min(512, sequence.sizes["z"])}},
             overwrite=False,
         )
-        logger.info("Acquisition completed successfully")
+
+
+def _enabled_fov_config(sequence: MDASequence) -> dict | None:
```
ieivanov: The functions `_enabled_fov_config`, `_build_prescan_sequence`, `_build_timelapse_sequence`, `_row_index_to_letter`, and `_filter_good_positions` should be within the `fov_selection` module or maybe methods of the `FovSelection` class, not within mantis_engine.py

shrimpy/mantis/mantis_engine.py
```diff
@@ -472,26 +527,191 @@ def acquire(
...
-                meta_path = data_path / "summary_metadata.json"
-                meta_path.write_text(json.dumps(to_builtins(meta)))
+        fov_cfg = _enabled_fov_config(sequence)
+        if fov_cfg is None:
+            logger.info(f"Starting acquisition: {name}")
+            self._run_mda(sequence, data_path, write_summary=True)
+            logger.info("Acquisition completed successfully")
+            return
+
+        self._acquire_with_fov_selection(sequence, output_dir, name, data_path, fov_cfg)
```
ieivanov: For clarity, I think it's better to not factor this into it's own function (`self._acquire_with_fov_selection`). I think the code will read cleaner with something like:

```python
if not _enabled_fov_config(sequence):
    # acquire regular MDA
    ...
    self._run_mda(sequence, data_path, write_summary=True)
else:
    # acquire MDA with fov selection
    ...
    # acquire pre-scan
    self._run_mda(prescan_seq, prescan_output, write_summary=prescan_output is not None)
    
    # build new MDA sequence and run acquisition
    good = list(self._fov_good_names)
    timelapse_seq = _build_timelapse_sequence(sequence, good)
    self._run_mda(timelapse_seq, data_path, write_summary=True)
```

shrimpy/mantis/mantis_engine.py
```diff
@@ -472,26 +527,191 @@ def acquire(
...
+def _enabled_fov_config(sequence: MDASequence) -> dict | None:
+    """Return the ``fov_selection`` metadata block when enabled, else ``None``."""
+    meta = sequence.metadata.get("mantis", {}) if sequence.metadata else {}
+    fov_cfg = meta.get("fov_selection")
+    if fov_cfg and fov_cfg.get("enabled"):
+        return fov_cfg
+    return None
+
+
+def _build_prescan_sequence(sequence: MDASequence, input_channel: str) -> MDASequence:
```
ieivanov: I still think it's better is pre-scan is configured with it's own valid and independent MDASequence, like I suggested in https://github.com/czbiohub-sf/shrimPy/pull/291#discussion_r3548012534. This will allow us to use a different z-plan - for example use a 2D scan rather than 3D, and maybe even acquire a timelapse and keep circling until the FOVs become "good" - for example we can start with low cell density and wait for the cells to grow on the microscope before we start acquiring data, or start with recently infected cells and wait until the cells start showing a phenotype before we start acquiring data. Some of these ideas will require some more work - for example terminating the pre-scan MDA once we've collected sufficient "good" FOVs, but I think it's good to seed this by using a proper MDASequence rather than hand-building it as in `return sequence.replace(channels=channels, time_plan={"loops": 1, "interval": 0})`

ieivanov: This MDA sequence can be nested under mantis:fov_selection:prescan_mda or provided with its own yaml file with `shrimpy acquire mantis -c regular_mda_config.yml --pre-scan-config prescan_config.yml`. I lean towards the first option. In both of these cases, in `_build_timelapse_sequence` we'll still call `sequence.replace(stage_positions=good_positions, metadata=meta)` - the starting stage_positions of the regular MDA can be empty.

ieivanov:
Looks great! I think this is very usable. My comments are somewhat nitpicky at this point. Looking forward to try this on mantis on Monday!

y222gu:
Thank you for the valuable feedback! Quick recap of what's changed over the last two rounds:

- **Pre-scan as its own MDA sequence:** Pre-scan is now a separate Multi-Dimensional Acquisition from the timelapse, so we can use different z-plans and time-plans for each. That allows us pre-scan in 2D while running the timelapse in 3D.
- **Parallel processing:** Pre-scan frame processing now runs in parallel with acquisition, which should cut down the time spent waiting on FOV selection decisions.
- **FOV selection module:** Pulled all the FOV selection helper functions out of `mantis_engine.py` into their own module. The code's a lot more organized now.
- **GPU optimization:** All preprocessing now runs on GPU, and it will raise an error if no GPU is detected or if something falls back to CPU. Also wrote a local Tensor-based version of flatfield preprocessing so it can run on GPU too.
- **Decision time estimate:** It takes about 15 seconds to decide on a Z stack, but that's based on Bruno's interactive-node GPU. I will verify this on Mantis's actual computer again.
- **Config file:** Users can now define a grid to pick which wells to scan, and pre-scan + timelapse parameters are now combined into one config file. One place to manage settings.
- **Saving decision-making outputs:** Added flags to save the pre-scan Z stack as .ome.zarr, plus all the decision-making products (preprocessed FOVs as PNGs, extracted features, and selection decisions in a CSV).
- **Better debugging:** Added logging for every preprocessing step (success/failure), which should make tracking down issues much easier.
- **Full acquisition test:** Ran a full test with the replay camera on 147 FOVs, no errors. System's holding up well.

y222gu:
This branch has been ported to #296.

ieivanov:
Superseded by #296

PR #293, opened 2026-08-06
Validate acquisition config metadata with ShrimpyMetadata

ieivanov:
**Stacked PR 1/2** — base of the stack, targets `main`.
Followed by #294 (`base_engine` → this branch).

## What

Fold the microscope settings directly into `MDASequence.metadata` (`autofocus`, `reset_hardware_sequencing_settings`, `dynatrack`) instead of nesting them under `metadata.mantis`, and validate them with a pydantic model so a mistyped setting fails before any hardware is touched.

```yaml
setup: ...            # ROI, imaging path, device properties applied once
channels: ...
metadata:
  autofocus: {enabled: true, method: PFS, stage: ZDrive}
  reset_hardware_sequencing_settings:
    - ['TS2_DAC03', 'Sequence', 'Off']
  dynatrack: {enabled: true, input_channel: BF, tracking_channel: BF}
```

## Changes

- add `shrimpy/config.py`: `ShrimpyMetadata` (strict, `extra="forbid"`) and `load_config()`, which parses the file and validates the metadata sections
- `MantisEngine` reads the validated metadata in `setup_sequence` / `teardown_sequence`
- add `DynaTrack.from_config()` for engines holding a `DynaTrackConfig`; keep `from_metadata()` as a dict wrapper
- update the mantis configs and the demo test artifact; drop the unused `pymmcore_widgets` metadata block
- note in `CLAUDE.md` that `mantis_acquisition_widget.py` is deprecated and still uses the old layout

## Compatibility

Configs using the older `metadata.mantis` nesting are rejected by `load_config()` with a migration message. The deprecated GUI widget still writes the old layout and is not migrated.

## Testing

`make test` (all green) — `shrimpy/tests/test_config.py` covers validation and the legacy-layout error; engine and DynaTrack tests updated to the new layout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #294, opened 2026-08-06
Extract shared BaseEngine from MantisEngine

ieivanov:
**Stacked PR 2/2** — targets `shrimpy_config_metadata_sections` (#293), not `main`. Review #293 first; the diff here is just the `BaseEngine` commit.

## What

Add `shrimpy/base_engine.py` with `BaseEngine`, the `MDAEngine` subclass shared by all microscope engines (`MantisEngine` today; iSIM and Dragonfly next). Mantis functionality is unchanged — this is a pure extract.

`BaseEngine` owns the platform-independent behavior:

- hardware sequencing defaults (`use_hardware_sequencing=True`, `force_set_xy_position=False`) and registration with `mmc.mda`
- debug logging of property changes, ROI changes, and XY stage moves
- autofocus: reads `metadata.autofocus`, dispatches to the simulated `demo-PFS` method or to the engine's `engage_autofocus()` hook, and raises `SkipEvent` when autofocus is enabled but does not engage
- protecting the autofocus stage from Z writes (`_set_event_properties`)
- resetting `metadata.reset_hardware_sequencing_settings` in `teardown_sequence`
- `acquire()` and the indexed acquisition naming

Microscope engines subclass it and override only what differs. `engage_autofocus()` raises `NotImplementedError` in the base; `MantisEngine` implements it with Nikon PFS.

`MantisEngine` keeps the mantis specifics: acquisition timeouts, XY stage speed modulation, DynaTrack, and Nikon PFS — down from ~520 to ~250 lines.

## Notable details

- The autofocus routines now **return a bool** instead of each mutating `_autofocus_success`; the base records it in one place.
- `MantisEngine`'s DynaTrack setup now runs **after** `super().setup_sequence()`, so the pixel size and any grid-plan FOV sizes reflect the state the setup event leaves the hardware in. No frames are produced during setup, so connecting `frameReady` later misses nothing.
- Removed a `logger.info("... setup completed successfully")` that fired before the setup event had actually been applied.

## Testing

`make test` — 285 passed. Shared engine tests moved to `shrimpy/tests/test_base_engine.py`, with added coverage for `_set_event_properties`, `_get_autofocus_z_position`, the logging callbacks, and the `NotImplementedError` path. The demo-core integration tests are otherwise untouched and still pass, which is the main evidence that mantis behavior is preserved.


ieivanov:
## Merge order: this one goes third

**Depends on #293**, which is its base and is currently `CONFLICTING` with `main` over `CLAUDE.md` (from #299). Once #293 lands, this PR merges **clean** — verified in the trial run, no conflicts at all.

Two housekeeping items before it can go in:

- it is still a **draft**;
- its base is `shrimpy_config_metadata_sections`, which will need retargeting to `main` once #293 merges.

Note that #296 is stacked on this branch and is by far the largest of the four (+15,985 across 49 files), so anything that changes here is cheapest to settle before that one starts its pre-merge sync.

ieivanov:
@jeskesen if you have a chance, it would be great to get a high-level review of this PR from you.

shrimpy/engines/base_engine.py
```diff
@@ -0,0 +1,569 @@
...
+        if self._autofocus_method == DEMO_PFS_METHOD:
+            self._autofocus_success = self._engage_demo_pfs(
+                event=event,
+                fail_at_index=self._autofocus_fail_at_index,
+            )
+            return
+
+        self._autofocus_success = bool(self.engage_autofocus(event))
+
+    def engage_autofocus(self, event: MDAEvent) -> bool:
```
jeskesen, line 333: Dragonfly & Mantis have a similar looping algorithm that appears to be substantially copy/pasted.  Might there be enough to warrant bringing it into the base class?

shrimpy/engines/base_engine.py
```diff
@@ -0,0 +1,569 @@
...
+        mmc.events.XYStagePositionChanged.connect(self._on_xy_stage_position_changed)
+
+    # ------------------------------------------------------------------
+    # Logging callbacks
+    # ------------------------------------------------------------------
+
+    def _on_property_changed(self, device: str, property_name: str, value: str) -> None:
+        """Log property changes at debug level."""
+        # Ignore select property changes
+        if property_name in ("PFS Status", "PFS in Range", "FocusMaintenance"):
```
jeskesen, line 123: This is Nikon specific...   It should be in the mantis class not the base?

shrimpy/engines/base_engine.py
```diff
@@ -0,0 +1,569 @@
...
+    def _get_autofocus_z_position(self, event: MDAEvent) -> float:
+        """Return the target Z position of the autofocus stage for ``event``.
+
+        Z positions are not written to the autofocus stage while autofocus is
+        enabled (see :meth:`_set_event_properties`), so the target position is
+        read from the event's properties when present, and from the stage's
+        current position otherwise. For a ``SequencedEvent``, the position of
+        the first frame of the burst is used.
+        """
+        event = first_event(event)
```
jeskesen, line 411: This could be a problem if an acquisition ever runs DynaTrack and autofocus on the same Z device, because DynaTrack writes its corrected Z onto the sequenced event while this reads the first sub-event — so the correction wouldn't be seen here and autofocus would engage at the uncorrected position.

No checked-in config hits it (dynatrack_demo.yaml pairs the devices but has autofocus disabled), so this is latent rather than broken. Worth a guard or a comment before someone enables both.

shrimpy/engines/mantis_engine.py
```diff
@@ -0,0 +1,175 @@
...
+        self._xy_stage_speed = None
+
+    def setup_event(self, event: MDAEvent) -> None:
+        """Prepare mantis hardware for each event.
+
+        Mantis acquires with hardware sequencing, so ``event`` is normally a
+        ``SequencedEvent`` covering a whole Z-stack and PFS is engaged once per
+        burst by :meth:`BaseEngine.setup_event`.
+        """
+        # TODO: debug resetting xy stage speed
```
jeskesen, line 61: Does this todo need to be done before merging?  I don't see a call to _adjust_xy_stage_speed anywhere.

shrimpy/engines/dragonfly_engine.py
```diff
@@ -0,0 +1,85 @@
...
+        core = self.mmcore
+        z_offsets = [0, -10, 10, -20, 20, -30, 30]  # in um
+
+        for z_offset in z_offsets:
+            core.setPosition(z_stage_name, z_position + z_offset)
+            core.waitForDevice(z_stage_name)
+
+            try:
+                core.fullFocus()
+            except Exception:
```
jeskesen, line 74: This could be a problem if AFC locks successfully but then fails to release the hold, because the adapter does that release after focusing — so the exception arrives on an already-good focus, and the code treats it as a miss, moves the stage to the next offset, and ultimately skips the frame. Rare, but the response is backwards. Checking whether a lock is actually held before deciding it failed would distinguish the two.

Separately: each failed attempt costs ~5 s inside the adapter, so a full offset search stalls the acquisition ~35 s with only debug-level output — worth logging the retries at INFO.

jeskesen (approved):
This looks good to me.  Thre are a couple of things I'd say should be addressed before before merging, but I'm going to "approve" so that I don't block your merge.  Happy to chat about any of the comments.

ieivanov:
Thanks @jeskesen! That's exactly the review I was looking for. I think these make sense, I'll look thru them in more detail before merging.

ieivanov:
TODO: add RobustMMCore to DrigonflyEngine

shrimpy/engines/base_engine.py
ieivanov: That's true - for now they have small but meaningful differences. I'll look for a way to unify the two autofocus mechanisms in the future.

shrimpy/engines/base_engine.py
ieivanov: Yes! Thanks, implemented.

shrimpy/engines/base_engine.py
ieivanov: True, added a check

shrimpy/engines/mantis_engine.py
ieivanov: Right, for now we don't use that, but we might in the future. I'll leave it as is. I seem to have forgotten what the bug actually is, we'll have to rediscover it

shrimpy/engines/dragonfly_engine.py
ieivanov: Good ideas, this method was a draft for now, I'll implement them in #304

PR #290, opened 2026-07-01
DynaTrack v2.0 refactor

ieivanov:
## DynaTrack v2.0 refactor

Supersedes #257.

Restructures DynaTrack from a mantis-only feature into a reusable, microscope-agnostic package with a validated, well-organized config, and modernizes the virtual-staining path for the `viscy` → `cytoland` split.

### 1. New `shrimpy/dynatrack/` package

Moved out of `shrimpy/mantis/` (via `git mv`, history preserved):

| New | Was |
|---|---|
| `dynatrack/tracking.py` | `mantis/dynatrack.py` — config + tracking algorithms + `DynaTrackUpdater` |
| `dynatrack/worker.py` | `mantis/dynatrack_worker.py` — GPU/torch worker subprocess |
| `dynatrack/preprocessing.py` | `mantis/dynatrack_preprocessing.py` — deskew → phase → VS pipeline |
| `dynatrack/position_update.py` | `mantis/position_update.py` — now package-internal |
| `dynatrack/manager.py` | **new** — `DynaTrack` coordinator (the only class engines touch) |
| `dynatrack/README.md` | **new** — algorithm + tracking-method reference |

`shrimpy/mantis/replay_camera.py` → `shrimpy/replay_camera.py` (shared by mantis + isim).

### 2. PositionUpdater folded into DynaTrack

`mantis_engine` no longer knows position-updaters exist. The new `DynaTrack` facade owns the position store, async update manager, per-position frame buffering, and the worker-subprocess lifecycle. `MantisEngine.setup_sequence` shrank from ~130 lines of wiring to a `DynaTrack.from_metadata(...)` call plus a `frameReady` connect. `PositionUpdateConfig` was deleted; custom / in-process trackers remain available via `DynaTrack(config, sequence, updater=...)`.

The unit boundaries are clean: `PositionUpdater` is a pure µm↔µm (stage-coordinate) transformer that never sees pixels; `DynaTrackUpdater` measures drift in pixels internally and converts to µm; `DynaTrack` derives and supplies the scales.

### 3. Validated, organized config

`position_update` is gone — DynaTrack settings live directly under `metadata.mantis.dynatrack`. `DynaTrackConfig` is now a **pydantic `BaseModel`** (`extra="forbid"`, so typos fail fast), with the tuning fields grouped into cohesive nested sub-models:

- **`shift`**: `maximum` (FFT search range), `limits` (per-axis deadband + clip), `dampening`
[...]

PR #296, opened 2026-08-11
Smart FOV selection on the BaseEngine refactor

### LINKED ISSUE #17: add FOV selection based on cell distribution
Develop methods for 1) predicting nuclei from BF images on the fly and 2) selecting FOVs based on the nuclei predictions. Most of the infrastructure for 2) can be ported from the `dragonfly_automation` package.

ieivanov:
Supersedes #291, which targeted `256-refactor` from `smart-fov-selection` directly. That base has since been absorbed here: this branch merges `256-refactor`, then `main`, then `base_engine`, so it carries everything #291 did plus the port onto the new engine layout. #291 can be closed once this lands.

Merges `smart-fov-selection` into the `base_engine` line and ports the FOV-selection pipeline onto the new engine layout. Opened as a draft — see **Open questions** below.

## What's here

Five commits on top of `smart-fov-selection`:

| Commit | What |
|---|---|
| `c455b0f` | Merge `256-refactor` — out-of-process napari viewer (#289), Python 3.12 (#292) |
| `ef47d55` | Merge `main` — DynaTrack v2.0 refactor (#290) |
| `fa9eb8d` | Pin acquire-zarr 0.9.0; always index the acquisition name |
| `5393871` | Merge `base_engine` — engine restructure, FOV port, int plate coords |
| `cb20e30` | Deduplicate preprocessing onto `shrimpy/preprocessing.py` |

Merging in that order kept each step small: by the time `base_engine` was merged, #289/#290/#292 were already in, so the only real conflict left was the engine restructure itself.

## The FOV-selection port

`base_engine` deletes `shrimpy/mantis/mantis_engine.py` and splits it into `shrimpy/engines/{base,mantis,isim,dragonfly}_engine.py`. The ~490 lines of FOV wiring moved into `engines/mantis_engine.py`:

- `_fov` / `_fov_passed_names` / `_fov_calibration_csv` / `_run_index` state
- `setup_sequence` — coordinator construction and worker start, sharing `_zyx_shape()` with DynaTrack
- `teardown_sequence` — selection capture for both the calibration and standard paths
- `acquire()` — the two-run adaptive acquisition, plus `_run_mda`, `_save_selected_fov_config`, `_launch_feature_viewer`
- the artifact-aware `_get_next_acquisition_name` / `acquisition_artifact_paths`

## Conventions adopted from `base_engine`

**Metadata flattening.** `base_engine` folds microscope settings directly into `metadata` (no `metadata.mantis` level). `fov_selection` now follows: `sequences.py` reads `metadata.fov_selection`, `ShrimpyMetadata` declares the section, and the three FOV configs are flattened. All 8 configs under `config/mda/` validate through `load_config`.

**Integer plate coordinates.** `plate_row`/`plate_col` are zero-based ints throughout, matching useq-schema `main`.

Worth recording: the `add-plate-row-col` branch (which allowed `int | str` coords and auto-generated position names) is **not** merged into useq-schema `main` — main took a narrower `int | None` approach. Main does accept int coords next to a field-suffixed name, so the string-label conversion in `expand_candidate_fovs` / `_filter_good_positions` is no longer needed and is dropped. One wrinkle: useq only auto-names positions it expands itself, so `expand_candidate_fovs` now derives an unnamed plate center's base name from its well label to keep `B2_0000` naming.

## Smaller changes

- `acquire-zarr` pinned to `0.9.0`. The old `==0.6.0` pin carried a *"even simple acquisitions would fail"* note; that does not reproduce on 0.9.0 (verified below).
- `_get_next_acquisition_name` always appends the index (`acq_1`, `acq_2`, ...) rather than using the bare name when free, matching `main`. This also fixed three `test_mantis_integration` failures that globbed for `test_acq_*.ome.zarr`.
- `config.py` accepts a `pymmcore_widgets` section (the version stamp `MDAWidget` writes) so a GUI-saved config isn't rejected by `extra="forbid"`.
- `uv.lock` regenerated with `uv lock` at each merge rather than hand-merged.

## Verification

A demo acquisition driven through the new engine layout writes a readable store:

```
stores: ['acq_1.ome.zarr']
zarr v3 / ngff 0.5, 12 positions, shape (3, 2, 13, 256, 1600)
well paths: ['A/1/fov0', 'A/1/fov1', 'A/1/fov2', 'A/1/fov3']
```

**Tests: 308 passed, 8 failed, 6 skipped** (227 passing before this work).

None of the 8 failures are introduced by these merges:

- **8 in `test_fov_selection_manager.py`** — stale tests against renamed APIs (`good_position_names` → `passed_position_names`, `_calibration_mode` vs `calibration_mode`, a changed `finalize` signature). Pre-existing on `smart-fov-selection`; they fail identically on that branch untouched.

Two test files were also excluded from the local run (`test_fov_selection_instanseg.py`, `test_fov_selection_pipeline.py`) because the optional `fov` group isn't installed in this environment — CI should cover them.

## Preprocessing deduplication

`smart-fov-selection` had unified DynaTrack's and FOV selection's label-free preprocessing into one shared `shrimpy/preprocessing.py`; #290 re-split it into `shrimpy/dynatrack/preprocessing.py`, and the merge left both. `cb20e30` restores the unification — the DynaTrack copy is deleted and its worker points at the shared module.

The shared module is a strict superset: the deskew / phase / VS implementations are identical, and it additionally offers a `flatfield` step (biahub `mantis-v2.nf` 0-flatfield, in torch so it runs on the GPU), `require_gpu` fail-fast, per-step labelled logging (`[<fov>] deskew ok (0.4s)`, preserving the timings the copy logged inline), `return_intermediates`, and `RECON_STEPS`.

That last point is a **behavior fix for DynaTrack**, not just a refactor: its copy returned `None` (with a warning) unless `deskew` or `phase` was present, so a `vs`-only or `flatfield`-only pipeline silently skipped reconstruction and tracked on the raw stack. Covered by a new regression test.

The call contract is unchanged — `DynaTrackUpdater` already declared `preprocessor: Callable[[np.ndarray], dict[str, torch.Tensor]]` and selects its channel by name, which is exactly what the shared `__call__` returns. Only the deleted copy's `build_preprocessor` return annotation said `-> np.ndarray`, and that was already stale.

## Open questions

1. **`test_channels_type_invalid_raises` was passing for the wrong reason.** The shared `META` fixture omitted the required `model.type`, so those tests raised on *that* instead of what they asserted. Adding it fixed 6 tests and exposed this one: the `channels_type` validation it checks never actually fires. That looks like a real gap in `FovSelection.from_metadata` rather than test debt.

2. **`test_base_engine.py` fix.** The Dragonfly exhausted-autofocus case asserted 7 `setPosition` calls; the engine makes 7 offset attempts plus a final restore to the target Z, so 8. Corrected here — flagging since it's `base_engine`'s own test.

3. **`origin/smart-fov-selection` has moved.** It carries one commit this branch doesn't (`f8c94a8 update for dragonfly test`). Say the word and I'll merge it in.


ieivanov:
## Preprocessing deduplication (`cb20e30`)

`shrimpy/dynatrack/preprocessing.py` is deleted; DynaTrack's worker now calls `shrimpy.preprocessing.build_preprocessor` with the explicit-kwargs signature, exactly as the original `smart-fov-selection` branch had it. Docstrings in `tracking.py` and the DynaTrack README were repointed.

I checked the shared module is a genuine superset before deleting the copy — the deskew / phase / VS implementations diff clean, and the shared one adds:

- a `flatfield` step (biahub `mantis-v2.nf` 0-flatfield, in torch so it runs on the GPU)
- `require_gpu` fail-fast, used by FOV selection
- `return_intermediates` for debug persistence
- `RECON_STEPS`

The only thing the copy logged that the shared one doesn't log inline is per-step timings, and the `_step` wrapper preserves those at INFO (`[<fov>] deskew ok (0.4s)`).

### This is a behavior fix for DynaTrack, not just a refactor

The private copy returned `None` (with a warning) unless `deskew` or `phase` was in the pipeline, so a `vs`-only or `flatfield`-only pipeline **silently skipped reconstruction and tracked on the raw stack**. The shared module builds a preprocessor for any reconstruction step. Added a regression test for it.

### The call contract needed no adaptation

`DynaTrackUpdater` already declared `preprocessor: Callable[[np.ndarray], dict[str, torch.Tensor]]` and selects its channel by name, which is what the shared `__call__` returns. Only the deleted copy's `build_preprocessor` annotation claimed `-> np.ndarray`, and that was already stale relative to its own `__call__`.

### Test fix

`test_pipeline_without_reconstruction_step_returns_none` was asserting the old DynaTrack-only semantics (`build_preprocessor(ZYX, ["vs"]) is None`) and failing. It now covers downstream-only pipelines, with the recon-step cases in the new test. That test also had a dead `records` assertion left over from a removed logging check.

### Verification

**308 passed, 8 failed, 6 skipped** — up from 306 / 9. The `waveorder` failure is gone (that test no longer reaches the import). The remaining 8 are the pre-existing stale-API tests in `test_fov_selection_manager.py`.

Re-verified end to end: a demo acquisition through `MantisEngine` writes `acq_1.ome.zarr`, 12 positions, shape `(3, 2, 13, 256, 1600)`.

---

Remaining open questions in the PR description: the `channels_type` validation gap, the `test_base_engine.py` Dragonfly fix, and the unmerged `f8c94a8` on `origin/smart-fov-selection`.

ieivanov:
## Merge order: this one goes last — and should sync with `main` first

**Do not merge this directly.** Merge `main` down into `smart-fov-selection-base-engine` first, resolve there, then merge up.

### Why

This branch collides with #297 on `shrimpy/preprocessing.py` and `shrimpy/tests/test_preprocessing.py`. The collision is unavoidable in **either** order: this branch carries the shared-preprocessing refactor but **not** #297's `use_waveorder` commit, so the two copies of `preprocessing.py` differ by 38 lines.

The only question is where that resolution happens. Trial merges:

| | merge #296 directly | sync first, then merge |
| --- | --- | --- |
| Where the collision lands | inside a **49-file, +15,985** diff | in the feature branch, in isolation |
| Test suite while resolving | not running | running |

A botched `preprocessing.py` resolution is easy to miss in the first column. The sync is worth doing regardless — the branch is 2 commits behind `main` and needs to absorb #295 and #299 anyway.

### Watch for this during the sync — git will NOT flag it

Merging after #295 auto-merges `pyproject.toml` with **no conflict** and produces two entries in `dependencies`:

```toml
"acquire-zarr>=0.9.0",   # from #295 on main
"acquire-zarr==0.9.0",   # from this branch
```

Duplicate PEP 508 requirements are ANDed, so this silently collapses #295's deliberate **floor** into an **exact pin**, quietly reverting its intent. Drop the `==` line and keep main's `>=`.

Also expect `uv.lock` conflicts at essentially every step — regenerate with `uv lock` rather than hand-merging.

### Before merging

- still a **draft**;
- base is `base_engine`, retarget to `main` once #294 lands;
- the 8 known-failing stale-API tests in `test_fov_selection_manager.py` (see description) are not merge blockers, but this is the natural point to clean them up;
- `89f5a6c disable autoshutter in DragonflyEngine` was pushed here after the description's verification run, so those numbers predate it.

ieivanov:
## Update: the `acquire-zarr` hazard is fixed on this branch

Confirmed we're keeping the **floor**, so `f24c03c` adopts main's line verbatim on this branch:

```toml
# arrives transitively via ome-writers, but declared directly to pin the floor
"acquire-zarr>=0.9.0",
```

Same text, same comment, same position in the `dependencies` list as `main` — the `dependencies` block is now byte-for-byte identical to main's, so the merge sees matching content and **cannot** reintroduce the silent duplicate described above. The old `acquire-zarr==0.9.0` pin is gone.

Two side effects worth noting:

- Relocking without the exact pin collapses three duplicate multi-version entries the pin was forcing the resolver to fork on (`numpy`, `scipy`, `shap`). That accounts for most of the `uv.lock` diff. No package is added or removed — 341 → 338 entries, `uv lock --check` clean, `acquire-zarr` still resolves to 0.9.0.
- Suite is unchanged: **383 passed / 8 failed**, the same known stale-API tests in `test_fov_selection_manager.py`. (383 rather than the 380 quoted earlier — `89f5a6c` added three.)

The rest of the pre-merge sync still stands: merge `main` down into this branch before merging up, and resolve the `preprocessing.py` / `test_preprocessing.py` collision with #297 there rather than inside this diff.

ieivanov:
## Update: #300 lands before all of these — two more conflicts to expect here

Revised order: **#300 → #297 → #293 → #294 → #296.** #300 merges clean onto `main` and adds nothing to #297/#293/#294; its entire cost lands on this PR, as two additional conflicts on top of the ones already noted:

| Step | previously | with #300 first |
| --- | --- | --- |
| +#296 | preprocessing.py, test_preprocessing.py, uv.lock | **+ pyproject.toml, base_engine.py** |

Both are small in substance. Resolve them during the pre-merge sync, along with the preprocessing collision.

### 1. `pyproject.toml` — keep `main`

```
HEAD:  pymmcore-plus = { ... branch = "main" }   <- keep (#300)
#296:  pymmcore-plus = { ... branch = "dev" }
```

Keep #300's `main`, and keep this branch's adjacent plate-coords comment.

### 2. `shrimpy/engines/base_engine.py` — looks big, is small

The conflict renders as #300's two-line replacement comment versus **this branch's entire `acquire()` + `_run_mda`**. That framing is misleading. What actually happened: #300 removed the `summary_metadata.json` workaround from `shrimpy/mantis/mantis_engine.py`; #294 had already copied that workaround into `base_engine.py`, and this branch moved it again into `_run_mda`. So the deletion needs re-applying at its new home.

Resolution: **keep all of this branch's code**, then inside `_run_mda` delete

```python
# Write summary metadata after the zarr store is created.
# TODO: remove once ome-writers supports root-level metadata natively.
if write_summary and output is not None:
    def _write_summary_metadata(...):
        ...
    self.mmcore.mda.events.sequenceStarted.connect(_write_summary_metadata)
```

replacing it with #300's comment:

```python
# Summary metadata is written by pymmcore-plus into the zarr root group's
# attributes, under `attributes.pymmcore_plus.summary_metadata`.
```

Then drop the now-unused `json` / `to_builtins` imports, and consider whether `_run_mda`'s `write_summary` parameter still earns its keep.

### CI will catch a bad resolution

`shrimpy/tests/test_mantis_integration.py` auto-merges cleanly and keeps #300's new assertions (reading `zarr.json` attrs instead of `summary_metadata.json`). So if the workaround is kept by mistake, `test_summary_metadata_written_to_zarr` fails rather than silently regressing #300.

---

Running list for the pre-merge sync of this branch, now:

1. `preprocessing.py` + `test_preprocessing.py` — 38-line divergence with #297 (this branch lacks its `use_waveorder` commit)
2. `pyproject.toml` — keep `pymmcore-plus` on `main`
3. `base_engine.py` — re-apply #300's workaround removal in `_run_mda`
4. `uv.lock` — regenerate, don't hand-merge
5. ~~duplicate `acquire-zarr` declaration~~ — already fixed in `f24c03c`

y222gu:
## Smart FOV selection

Online, streaming selection of fields of view during an acquisition, plus a feature
viewer for tuning the selection model and offline scripts for training one. Microscope
agnostic and built on `BaseEngine`: an engine builds a `FovSelection` coordinator from its
`metadata.fov_selection` config and interacts with that object only.

### Workflow: two-run adaptive acquisition

1. **Pre-scan**: image every candidate FOV once on `fov_selection_channel`, score each,
   decide which pass.
2. **Timelapse**: the real multi-timepoint acquisition, on the selected FOVs only.

**Calibration mode** stops after the pre-scan, extracts every feature, and opens the feature
viewer so you can design a model before restart for a full run.

### Per-FOV decision pipeline (`pipeline.decide_fov`, in a worker subprocess)

```
raw z-stack
  # preprocessor (preprocessing.py): reconstruction only
  -> reconstruct   deskew / phase / virtual staining  -> {channel: (Z, Y, X)}
  # FOV-selection decision (this package):
  -> project       sum | max | middle | logstd | best_focus_z
  -> reduce to ONE 2D image by `target` (cells | nuclei)
  -> segment ONE mask   cellpose | instanseg | otsu
  -> features           per-object -> FOV-level (plain single-mask keys)
  -> verdict            model.predict(features)
```

Projection/segmentation are kept out of the preprocessor: reconstruction is shared with
DynaTrack and is the natural 3D-to-2D boundary, so `preprocessing.py` stays free of selection
concerns.

### Structure

```
manager.py               engine-facing coordinator: buffering, verdicts, drain, selection
sequences.py             build the pre-scan / timelapse MDASequences
pipeline.py              per-FOV decision (project -> segment -> features -> verdict)
worker.py                subprocess isolation (torch/GPU)
fov_model.py             pluggable models + interpretable curve conversions
segmentation.py          Cellpose / InstanSeg / Otsu backends
feature_extraction.py    object-level, FOV-level, and mask-derived features
prescan_artifacts.py     per-FOV PNG / CSV / OME-Zarr writers + finalize
acquisition_artifacts.py once-per-run records (recovery config, viewer launch)
feature_viewer/          Qt GUI
```

### Models

`model.type`: `ranking_by_defined_range` (per-feature desirability curves, keeps the top
`top_fov` per position; `top_fov` required), `classification_by_thresholding` (hard `[lo, hi]`
box), `classification_tree` (trained `.joblib`). Selection is by model type; `threshold` is
classification-only.

### Artifacts

All outputs land next to the acquisition; debug outputs are under `<name>_fov_debug/`, gated by
flags (`calibration_mode` forces `save_decision` on). The image folders and `fov_summary.csv`
name are the same in both modes, so a run loads in the viewer either way.

- `prescan_fov/`, `prescan_mask/`: per-FOV projection and mask-overlay PNGs (`save_decision`).
- `fov_summary.csv` holds one row per FOV: `filename`, features, `well_row`/`well_col` (both
  modes); normal mode also stamps `selected`/`position`/`rank`. This is the file the viewer loads.
- `selected_fov/`: the selected FOVs' projections, gathered after the drain (normal mode).
- `<name>_prescan.ome.zarr`: full per-step reconstruction (`save_pre_scan_omezarr`).
- `<stem>_best_focus_z.csv`: detected best-focus slice/depth (`save_best_focus_z_for_debug`).
- **`config_for_recovery.yaml`**: written once per normal run between the two passes. It is the
  acquisition config with the SELECTED FOVs filled into `stage_positions` (absolute x/y, the
  well's coarse ZDrive focus, plate row/col). It is a ready-to-run config that reacquires
  exactly the chosen FOVs, so a crashed or extended timelapse can be resumed without re-running
  the pre-scan. A `_<run_index>` suffix is added only if the name is already taken.

### Feature viewer

`python -m shrimpy.fov_selection.feature_viewer [CSV ...]` opens a Qt GUI to explore features
and tune the selection model, each FOV shown as a thumbnail. In **calibration mode** it
launches automatically on the pre-scan's `fov_summary.csv`, seeded with the model from the
acquisition config. You adjust the per-feature desirability profile, then **Save** it to a
profile YAML and copy that `features:` block into a full acquisition config's `model` section
to drive a real (non-calibration) run.

### Config, logging, testing

- All settings live under `metadata.fov_selection`, validated by `FovSelection.from_metadata`
  before any hardware is touched. See `config/mda/fov_selection_demo.yaml`.
- Logging: console at INFO (setup, per-FOV step status + one score line, a post-drain selection
  summary), file at DEBUG for detail; identical in both modes.
- Full test suite passes (`test_fov_selection_*`, `test_fov_model.py`, `test_base_engine.py`).

ieivanov:
TODO:
 - [ ] remove `fov_selection_integration_plan.md` and `replay_camera_todo.md` and open issues for outstanding features

ieivanov:
Ready to merge after a fresh test on mantis

PR #297, opened 2026-08-11
Extract label-free preprocessing into shrimpy/preprocessing.py

ieivanov:
## Why

The deskew → phase → virtual-staining pipeline is **general label-free reconstruction**, not something specific to position tracking. It currently lives in `shrimpy/dynatrack/preprocessing.py` and is coupled to `DynaTrackConfig`, so DynaTrack is the only thing that can use it.

Smart FOV selection needs exactly the same reconstruction: its pre-scan turns a brightfield z-stack into virtually-stained channels before segmenting and scoring each candidate FOV. Reaching into `shrimpy.dynatrack` from `shrimpy.fov_selection` to get it would make one feature depend on another for no reason, and copying it would leave two implementations to keep in sync.

So this moves the module to the package root and decouples it from `DynaTrackConfig`. DynaTrack keeps working exactly as before; FOV selection (#296) then imports the same module rather than carrying a second copy.

**Nothing about the reconstruction itself changes.** The deskew / phase / VS implementations are moved verbatim.

## Reviewing this

Three commits, each green on its own:

| Commit | | Tests |
| --- | --- | --- |
| `098feb5` | **Pure `git mv`** — module byte-for-byte identical, a 100%-similarity rename with a 0-line diff. Only the three references to it are repointed. | 247 (= `main`) |
| `c09f0cc` | Decouple from `DynaTrackConfig`; new opt-in options; tests. | 258 |
| `c9cf3a6` | Import waveorder only when phase reconstruction runs. | 261 |

The first commit is the whole move, so you can confirm it changes nothing and then read the second as an ordinary diff. Squashed together the file only matches at 43% similarity — below git's default rename threshold — which makes the move impossible to review.

Every step that transforms pixels is unchanged: `_deskew`, `_reconstruct_phase` and `_compute_transfer_function` are byte-identical, and `_predict_vs` / `_load_vs_model` / `_log_gpu_memory` differ only in log wording (`"DynaTrack: ..."` → `"Preprocessing: ..."`). With docstrings and comments stripped, 128 of `main`'s 171 code lines (74%) survive verbatim; the rest is plumbing.

## What changes

`shrimpy/dynatrack/preprocessing.py` → `shrimpy/preprocessing.py`.

`build_preprocessor` takes the reconstruction settings as explicit arguments rather than a config object:

```python
build_preprocessor(
    zyx_shape, preprocessing,
    deskew=..., phase=..., virtual_staining=..., output_channel=...,
)
```

The worker maps its config onto them. Note `output_channel` comes from `tracking_channel`, not `input_channel` — for a VS pipeline it names the VS target the updater tracks.

`DynaTrackUpdater` is untouched: it already declared `preprocessor: Callable[[np.ndarray], dict[str, torch.Tensor]]` and selects its channel by name, which is what `__call__` returns. Only the old `build_preprocessor` return annotation said `-> np.ndarray`, and that was already stale relative to its own `__call__`.

### Behavior changes

Three, all widening what's accepted — an existing DynaTrack config behaves identically.

1. **Any reconstruction step now builds a preprocessor.** The old code returned `None` (with a warning) unless `deskew` or `phase` was present, so a `vs`-only or `flatfield`-only pipeline silently skipped reconstruction and tracked on the raw stack.

2. **waveorder is imported only when phase reconstruction runs.** `warm_up()` resolved the compute device through `waveorder.device.resolve_device` unconditionally, so building *any* preprocessor imported waveorder — even a pipeline that never calls into it (flat-field is pure torch, deskew is biahub, VS is cytoland). On an install that legitimately lacks waveorder that raised `ModuleNotFoundError`. `_resolve_device` now takes `use_waveorder`, and `warm_up` passes `_needs_waveorder`, true only when phase settings are configured. The phase path still defers to waveorder so the transfer function and the inverse land on the device it picked.

   The torch fallback is CUDA-or-CPU, deliberately **not** MPS: the steps it covers are biahub's deskew and cytoland's VS, neither tested against the Metal backend, and silently moving a Mac run onto MPS would trade a working CPU run for an unsupported-operator failure deep in a dependency.

3. **New opt-in options**, all defaulting to off: a `flatfield` step (biahub `mantis-v2.nf` 0-flatfield, in torch so it runs on the GPU), `require_gpu` fail-fast, and `return_intermediates` for debug persistence.

Per-step logging moved into a `_step` wrapper that prefixes an optional label and reports `[<label>] deskew ok (0.4s)`, preserving the timings the old code logged inline.

## Verification

Measured against `main` in the same environment. Worth noting: with a plain `uv sync`, `test_dynatrack.py` is **skipped wholesale** (module-level skip — torch is only in the optional `dynatrack` group), so the DynaTrack suite doesn't run at all by default. These numbers are with `uv sync --group test-cpu` so it actually does:

| | `main` | this branch |
| --- | --- | --- |
| Full suite | 247 passed | 261 passed, 1 skipped |
| DynaTrack suites | — | 112 passed |

The +14 are the new tests; no regressions. The 1 skip is the flat-field-vs-biahub reference comparison (biahub not installed locally).

Also exercised the real `_LabelfreePreprocessor.__call__` end to end on CPU with a flat-field pipeline, **with waveorder absent** — it builds, runs, produces the right channel key, `DynaTrackUpdater` reads it back, and `'waveorder' not in sys.modules`.

### Tests added

- `shrimpy/tests/test_preprocessing.py` — step selection, the channel-dict contract, `require_gpu`, the flat-field/biahub equivalence, and the waveorder decoupling.
- `TestWorkerPreprocessorWiring` in `test_dynatrack.py` — the config → kwargs mapping. This is the seam the refactor introduces and nothing else covers it: a field wired to the wrong argument would leave DynaTrack tracking on a mis-reconstructed volume, and every existing test would still pass.

## Relationship to #296

#296 (smart FOV selection on the BaseEngine refactor) is the consumer. It currently carries this same change as one commit among several; landing it here first keeps the refactor reviewable on its own and lets it be validated against DynaTrack in isolation, which is what these numbers are. #296 will merge this branch rather than duplicate it — the shared module, worker, and test files are byte-identical on both sides, so that merge is conflict-free.


ieivanov:
@tayllatheodoro can you please test this PR on your next DynaTrack run? It refactors `preprocessing.py` so that it can also be used for smart FOV selection.

ieivanov:
## Merge order: this one goes first

**Nothing blocks this PR.** It merges cleanly onto current `main`, it is the only one of the four that is not a draft, and it is deliberately small (3 commits, 7 files) and validated against `main` in isolation — see the verification table in the description.

Going first also means #296 can absorb it during its pre-merge sync, which is where the one real collision between them should be resolved (details in the note on #296).

PR #303, opened 2026-09-08
Bugfixes and lots of logging from df2

ieivanov:

PR #304, opened 2026-09-09
Dragonfly AFC fixes: per-position autofocus, focus homing, and a zero-distance move guard

ieivanov:
Fixes found while bringing up AFC autofocus on the Dragonfly. The debugging
instrumentation that found them has been removed again; what is left is the
behavior changes, in `BaseEngine` where they are shared and in
`DragonflyEngine` where they are Leica-specific.

Stacked on #294 (now merged), which extracted `BaseEngine`. Supersedes #303,
auto-closed when its head branch was renamed; the pre-rebase state is archived
at `archive/smart-fov-selection-base-engine-df2-fixes`.

## Autofocus engages once per position, not once per Z index

`_should_engage_autofocus` keyed on `index['z'] == 0`, which re-engages once
per *channel*: the channel axis is outer to Z, so a 4-channel stack at one XY
location presents `c=0,z=0`, `c=1,z=0`, … all of which are `z == 0`. It now
keys on the time/position/grid indices (`AUTOFOCUS_POSITION_AXES`), so the lock
obtained on arrival is reused for every channel and slice acquired there, and a
new timepoint, stage position or grid site re-engages.

A failed attempt records its position too, so the remaining frames of that
position do not each retry the hardware — `_autofocus_success` carries the
outcome and `setup_event` skips the whole position.

**Effect on mantis** - autofocus used to engage once per channel, and we would something see that, for example, autofocus fails on `c=0`, but engages on `c=1`. This behavior was not exactly desirable and will be eliminated now.

## Zero-distance focus-stage moves are skipped (`MIN_FOCUS_MOVE_UM`)

This is the fix for the intermittent `waitForDevice` timeout on the Dragonfly.
Commanding a Leica FocusDrive to the position it already holds *hangs*: the
LeicaDMI adapter derives `Busy()` from the asynchronous `$71004` z-drive motion
flag, and a move the firmware resolves to zero encoder steps never moves the
motor, so the flag never arrives, `Busy()` stays True, and `waitForDevice`
throws after the 5 s core timeout. Whether a nominally-zero move crosses an
encoder step boundary comes down to a few nm of dither, which made it fail on
~2% of engagements while the rest cleared in 0.000 s.

`_move_focus_stage()` now reports the stage as already on target when the
distance is under 0.1 µm — far below the depth of field of any objective on
this scope — and reports move failures instead of raising, so the caller tries
the next Z offset.

## Core-Focus homing, opt-in via `metadata.autofocus.home_focus_device`

On the Dragonfly the z_plan rides the ASI piezo (Core-Focus) while AFC drives
the Leica FocusDrive, so AFC was locking onto wherever the previous Z-stack
stopped — the top slice — and the whole stack was acquired below focus. The
Core-Focus device's home position is now captured after the setup event and
restored before autofocus engages at each new position, and once in teardown.

This is opt-in because Core-Focus differing from the autofocus stage does not
by itself mean the two interact. Mantis has that same shape — Core-Focus is the
light-sheet scan galvo, which moves neither sample nor objective, so PFS on
`ZDrive` is indifferent to it — and homing there would only add a redundant
galvo move before every sequenced burst. `config/mda/dragonfly/dragonfly.yaml`
opts in; the default is off.

## Also

- Autofocus settings are cleared when a sequence disables autofocus. One engine
  runs more than one sequence (FOV-selection pre-scan, then the timelapse), and
  settings that survived would silently autofocus a run that asked not to,
  using the previous run's stage and method.
- The Dragonfly AFC routine returns early when continuous focus is already
  locked.
- `config/mda/mantis/mantis.yaml` enables autofocus.

## Testing

`shrimpy/tests/test_base_engine.py` covers the position keying (including the
4-channel stack that the old `z == 0` rule mis-handled), the homing opt-in and
its per-run reset, the `MIN_FOCUS_MOVE_UM` threshold either side of the
boundary, and the move-failure paths. Full suite: 347 passed, 1 skipped.

Verified on the microscope: the figure-producing acquisitions this PR deploys
were collected with these fixes in place.


ieivanov:
Ready to merge after testing on mantis microscope

ieivanov (commented):
Tested on mantis microscope, OK to merge

PR #310, opened 2026-09-19
Read the live viewer's data from the acquisition's OME-Zarr

ieivanov:
Switches the live viewer from a shared-memory ring of recent frames to reading the acquisition's OME-Zarr store read-only, now that [acquire-zarr#234](https://github.com/acquire-project/acquire-zarr/pull/234) has lifted the `zarr.json` read lock. Follow-up to #289.

## What changes

No image data crosses a process boundary any more. The ring buffer, the frame queue and the per-frame `frameReady` callback on the acquisition thread are all gone; `LiveViewer` sends the child process exactly two messages, `open <path>` and `finish`.

```
src/shrimpy/viewer/
├── store.py            AcquisitionStore: read a live or finished OME-Zarr
├── _napari_process.py  the window; run_viewer() is also `shrimpy view`
└── live.py             LiveViewer: spawn it beside an acquisition
```

What this buys:

- **The whole acquisition is browsable while it runs**, not just the last N frames the ring held.
- **Nothing is lost during napari's startup.** Previously the opening frames were pushed at a viewer that was not draining yet; now they are simply on disk when it opens. In a demo run below, the acquisition finished in 1 s — before napari was ready — and the viewer still showed all three timepoints.
- **Live and offline are the same code path**, which is what #289 set out to converge on.
- New `shrimpy view PATH [--deskew]` uses that same entry point. Because the store can be opened while it is being written, it also **attaches to a run already in progress** — from another terminal, or another machine that can see the directory.

## AcquisitionStore

Layout, ordering, channel names, axes and scales come from `iohub.open_ome_zarr`, which already handles every layout ome-writers writes (single image, bioformats2raw series, HCS plate) and re-reads array metadata on each `Position.data` access. On top of that, the store adds what reading a *live* store needs.

**A decompressed-volume cache.** shrimPy chunks a whole z-stack into one chunk, so reading *any* single plane decompresses the entire stack. Measured on a 164 MB volume (200 z × 256 × 1600; a real mantis stack is ~5× that):

| | per plane |
|---|---|
| plain zarr indexing (what iohub hands you) | **275 ms, every time** |
| `AcquisitionStore`, volume cached | **~0 ms** (first touch ~290 ms) |

Without it, scrubbing z through a deskewed stack would be minutes of decompression. The cache is LRU under a byte budget (`AcquisitionStore(cache_mb=...)`, 4 GB by default).

**`volume_ready()`.** Chunks are flushed by a writer thread once complete, so "the frame was acquired" is not "the frame is readable", and an unflushed chunk reads back as zeros. Checking the chunk files on disk both blanks half-written stacks instead of deskewing garbage, and keeps a blank volume from being pinned in the cache for the rest of the session.

**Plus** positions whose array does not exist yet (acquire-zarr creates it on first write), reads past the acquired extent, and the `latest_complete` frontier the sliders follow.

## Deskew geometry is unchanged

Same values as before: the **scan step** comes from the store's z scale — the same number the old viewer took from `z_plan.step` — and the **lateral pixel size** stays on the `napari-deskew-preview` default, overridable in the widget.

I initially read the pixel size from the store too, and it rendered differently. The store records whatever `getPixelSizeUm()` resolved to, which depends on transient device state and has silently changed the deskew ratio between otherwise-identical runs before (see the comment in `BaseEngine._setup_dynatrack`). Reverted. The deskew *math* was verified bit-identical across the old ring gather, the new store gather, and the offline plugin's `array_gather`; a regression test now pins the live-vs-offline equality.

## Other

- `iohub` moves into the `viewer` extra; `shrimpy acquire` stays lean without it.
- `DeskewControls` gains `angle_deg` / `pixel_size_um` arguments so the panel can open on the acquisition's geometry.
- `BaseEngine.data_path` is now public, so a `sequenceStarted` listener can find the store.
- **`--napari-cache-mb` is removed**, and `shrimpy view` does not gain a `--cache-mb`. The flag sized the frame ring, so its value decided how much of the run stayed visible; now every plane is readable from disk at any index and the budget only trades RAM against re-decompressing a volume. It is `AcquisitionStore`'s default, tunable in code. (The flag also carried a stale 8192 MB default that never matched `DEFAULT_CACHE_MB`.)

## Verification

- 309 tests pass (33 new), `make check` clean. Viewer tests build real ome-writers/acquire-zarr stores rather than hand-rolled zarr — the behaviour under test *is* that writer's — and drive the layer logic through a fake napari viewer, so they need no GUI.
- Ran `shrimpy acquire mantis --napari-viewer` on the demo config end to end, and `shrimpy view` on a deskewed store and a 4-position plate. Screenshotted a live run following t=0→3 as the store grew, with deskew on and the scan step read from the store.


ieivanov:
Ready to merge after https://github.com/imaging-formats/yaozarrs/pull/58

